PR #10, opened 2018-08-03
Upload all NYC schools via a rake task

LucasCioffi:


db/migrate/20180802145655_add_address_to_schools.rb
```diff
@@ -0,0 +1,12 @@
...
+    add_column :schools, :address_line_1, :string
+    add_column :schools, :address_line_2, :string
+    add_column :schools, :state, :string
+    add_column :schools, :postal_code, :string
+    add_column :schools, :phone_number, :string
+    add_column :schools, :borough, :string
+    add_column :schools, :created_at, :datetime
+    add_column :schools, :updated_at, :datetime
+  end
+end
```
daryanypl: One more change:  currently, the z code is stored in the "code" column of the schools table, as a plain string.  We're planning to have a zcode lookup table, with the columns:  id (MLN primary key), sierra_id (Sierra primary key), code (the string zcode).  And then the code column of the schools table would become a foreign key.

This would allow us to send the correct numeric pcode4 id to Sierra API (the pcode4 corresponds to the Sierra id primary key, rather than to the zcode string).

.gitignore
```diff
@@ -23,6 +23,9 @@
 # Ignore .dump
 /mylibnyc.dump
 
+# Ignore imported CSVs
+/data
```
daryanypl: I'm concerned that I see a .csv in /data that got pushed and it part of this PR.  We might need to sanitize the repository again.  The good news, is that we won't have branch histories accumulating over time on top of this one.

db/schema.rb
```diff
@@ -56,7 +56,7 @@
     t.string   "primary_language"
     t.string   "call_number"
     t.text     "description"
-    t.text     "physical_description"
+    t.string   "physical_description"
```
daryanypl: For my edification, what's the difference between text and string in ActiveRecord?

db/schema.rb
```diff
@@ -159,7 +167,7 @@
     t.string   "availability"
     t.string   "isbn"
     t.string   "language"
-    t.string   "physical_description"
+    t.text     "physical_description"
```
daryanypl: But here you changed to text?  How does the text-string conversion work?

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
     end
   end
 
+  desc "Import all NYC schools"
+  # task import_all_nyc_schools: :environment do
+  task import_all_nyc_schools: :environment do
+    csv_text = File.read('data/2016_-_2017_School_Locations.csv')
```
daryanypl: I'm not 100% sure what's being done here, but it looks like maybe this is a script for ingesting a schools csv file.  If so, I'd like to specify the name of the csv file to ingest in the command line arguments, when you call the script.  If the script is called with wrong arguments, it should output a usage notification message.  We can then later call the script with on-demand arguments from a user-facing webpage.

I'd also like to make a short mock data file, with just a couple of schools in it, for the unit tests.

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
...
+    rows = CSV.parse(csv_text, headers: true)
+    rows.each do |row|
+      row_hash = row.to_hash
+      school = School.where(code: row_hash['ATS SYSTEM CODE'].strip).first_or_initialize
```
daryanypl: I am so excited that we're already writing code to import the csv.

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
...
+      school.name = school.name || row_hash['LOCATION_NAME'].strip
+      school.address_line_1 = school.address_line_1 || row_hash['PRIMARY_ADDRESS_LINE_1'].strip
+      school.state = school.state || row_hash['STATE_CODE'].strip
+      if row_hash['Location 1'].present?
+        school.address_line_2 = school.address_line_2 || row_hash['Location 1'].strip
+        school.borough = school.borough || borough(row_hash['Location 1'].strip)
```
daryanypl: Could we capitalize and trim the borough field?

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
...
+      school.phone_number = school.phone_number || row_hash['PRINCIPAL_PHONE_NUMBER'].strip
+      school.created_at = Time.zone.now if school.id.nil?
+      school.updated_at = Time.zone.now
+      school.save
+      puts school.name
+    end
+  end
+
+  def borough(borough_string)
+    if borough_string.include?('Manhattan')
```
daryanypl: Even though we're importing data that should be clean, how about we do these comparisons case- and whitespace-insensitive, and save them in caps to make future processing easier?  Not sure how the ruby would look, but something like "if borough_string.uppercase.trim.include('QUEENS') then 'QUEENS' ".

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
...
+      'Manhattan'
+    elsif borough_string.include?('Queens')
+      'Queens'
+    elsif borough_string.include?('Brooklyn')
+      'Brooklyn'
+    elsif borough_string.include?('Statten')
```
daryanypl: two t's

lib/tasks/ingest.rake
```diff
@@ -389,6 +389,44 @@ namespace :ingest do
...
+      'Statten Island'
+    elsif borough_string.include?('Bronx')
+      'Bronx'
+    end
+  end
+
   desc "Deactivate schools from authorized list"
   task :deactivate_schools, [] => :environment do |t, args|
     path = 'db/2015SchoolUpdate/dropped-schools.csv'
```
daryanypl: Hardcoded path.  It looks like old code, so we don't need to fix in this PR, but could we put a TODO marker on it to remember to fix later?

lib/tasks/ingest.rake
```diff
@@ -515,9 +553,9 @@ namespace :ingest do
         File.foreach(path) do |line|
           next if $. == 1
 
-          # This appears in one of the dumps: 
+          # This appears in one of the dumps:
           # "p55258827"|"151"|"EDWARDS, TESNA"|"27777016362867"|"zx080"|"TEDWARD2@SCHOOLS.NYC.GOV";"TEdward2@schools.nyc.gov"
```
daryanypl: Let's definitely delete the user's email info, and clean the history.  Sorry this is still a bother, but hopefully we'll be finding fewer such history-cleaning triggers in the future.  Maybe let's do a global search for any other email address mentions.  We want to do the cleaning before the commit histories get interwoven into branch histories.

For this comment doc string, which is useful in explaining the following bit of code, let's replace the user's real email and name with a mock, like "JANEDOE@SCHOOLS.NYC.GOV".

lib/tasks/ingest.rake
nnat425: In addition, why is Location 1 lower-case, while other hash key's are upper-case? Can we keep consistency and have everything upper-case?

lib/tasks/ingest.rake
```diff
@@ -486,11 +524,11 @@ namespace :ingest do
   end
 
 
- 
+
 =begin
```
nnat425: Can we delete commented lines from 528 to 648?

lib/tasks/ingest.rake
```diff
@@ -486,11 +524,11 @@ namespace :ingest do
   end
 
 
- 
+
```
nnat425: Can we delete commented lines from 528 to 648?

lib/tasks/ingest.rake
nnat425: Why are we harding coding 'data/2016_-_2017_School_Locations.csv' ?

db/migrate/20180802145655_add_address_to_schools.rb
```diff
@@ -0,0 +1,12 @@
+class AddAddressToSchools < ActiveRecord::Migration
```
nnat425: To clarify, we're adding address fields because in order to create a patron for the v2 route, it requires the teacher's school address. Am I understanding correctly?

nnat425:
I think it would be great if we created some documentation/updated the README in regards to how the ingestion process works and how a developer can do it.

db/migrate/20180802145655_add_address_to_schools.rb
LucasCioffi: Yes @nnat425 you will pass the teacher's school's address to the microservice as the teacher's address.

lib/tasks/ingest.rake
LucasCioffi: I'll take your suggestion @nnat425 and change it, but the original reason was that passing in the CSV name as a variable to the rake task doesn't save much time.

The reason is that to run this rake task, we'll need the CSV to be available locally, so we will have to do a production deploy whenever there are new schools to add.  When we add the CSV, the developer would change this line too.  Having the file name in the code helps document where this CSV looks for its data in the file tree, but I can add that as a comment instead.

lib/tasks/ingest.rake
LucasCioffi: @nnat425 Darya suggested creating a new Git repo once this change is rolled in.  Since you have more PRs in this repo, I think it makes sense for you to decide on the best timing.

lib/tasks/ingest.rake
LucasCioffi: We agree that @nnat425 it's not doing any harm keeping it in.  I suggest keeping it since 

This code represents some of Paul's thinking from five years ago and how they addressed the ingesting then.  Here's the Git blame: https://github.com/NYPL/MyLibraryNYC/blame/master/lib/tasks/ingest.rake#L489

Judging by how long this project is living, this thinking might become relevant again if we start importing those types of CSVs again.

README.md
```diff
@@ -27,6 +27,14 @@ To dump data to seed-data dump file (i.e. for use by other devs):
 pg_dump -U_postgres mylibrarynyc -f db/seed-data.sql --data-only --exclude-table="schema_migrations|holds|users" --inserts --column-inserts
 ```
 
+To Import New NYC Schools:
+
+Assuming your CSV is public data, upload it under root/data/public.
+
+Run this command from the console locally and on AWS (replace the filename in the command): `rake ingest:import_all_nyc_schools['data/public/2016_-_2017_School_Locations.csv']`
+
+If the school is not found by zcode, the rake task will create a new record.  If the school is found by zcode, it will overwrite school's name, address_line_1, state, address_line_2, borough, postal_code, and phone_number with the data in the CSV.
```
LucasCioffi: @nnat425 does this documentation look sufficient?  This PR is ready for another look.  Gracias!

db/schema.rb
LucasCioffi: @daryanypl the exact difference depends on the database (MySQL vs. Postgres).  String has a default and max of about 255 characters.  Text has a default and max of about 65,000 characters.  Apparently in Postgres there is no performance advantage: https://stackoverflow.com/questions/3354330/difference-between-string-and-text-in-rails

I removed this code change in a later commit when I noticed it, and you saw that commit which reverts this change.  So this PR as a whole doesn't make a change to the schema for this field.

I don't know the exact reason that my schema was different than what was in source control, but it's not alarming because the one that's in source has `text` rather than `string` so no data will be truncated.

db/schema.rb
LucasCioffi: @daryanypl in Rails, the Schema is useful for creating a database, and migrations are used for changing an existing database (ie converting a field type from string to text).  So changes to the schema have no affect on existing databases.  I made this code change to revert a change to this line on a previous commit.  So this PR as a whole doesn't make a change to this file.

lib/tasks/ingest.rake
LucasCioffi: @nnat425 the file comes this way from DOE.

@daryanypl we set the borough using a separate method which checks for the presence of the borough name in the string.  I added a line to raise an exception if the borough is not found.

lib/tasks/ingest.rake
LucasCioffi: @daryanypl I made the change so that we are capitalizing the boroughs.

LucasCioffi:
Thanks, @nnat425.  It's ready for another look.

PR #11, opened 2018-08-03
Generate barcode

daryanypl:
NOTE:  Provisional, have not yet tested on new codebase.


app/models/user.rb
```diff
@@ -89,15 +89,37 @@ def send_unsubscribe_notification_email
...
+    assign_barcode
+    send_request_to_patron_creator_service
+  end
+
+
+
+  def assign_barcode
+    Rails.logger.debug("assign_barcode: start")
+    last_user = User.where('barcode < 27777099999999').order(:barcode).last
+    self.barcode = last_user.barcode + 1
```
nnat425: Where are we adding a prefix of '27777' ? Need to add that.

test/integration/user_flow_test.rb
```diff
@@ -22,4 +22,11 @@ class UserFlowTest < ActionDispatch::IntegrationTest
     follow_redirect!
     assert 'div', 'Welcome! You have signed up successfully'
   end
+
+
+  # Test that generated barcode is a 14-digit string, starting with '27777'.
+  test "user method generate_barcode returns valid string" do
+    assert(/\A2777\d{10}\z/ === user.generate_barcode)
+  end
```
nnat425: Currently broken. Will fix in a later PR due to time constraints.

db/schema.rb
```diff
@@ -181,6 +181,7 @@
     t.integer  "total_copies"
     t.string   "primary_subject"
     t.string   "bnumber",                     :limit => 20
+    t.string   "test_column",                 :limit => 20
```
nnat425: We can get rid of this.

LucasCioffi: It keeps showing up in PRs, and Kevin mentioned we'll keep the field in production, so I think we should keep it in this PR so we don't keep seeing it in commits.

nnat425: That's fine, once this is merged, I'll create a PR to remove it. Thoughts?

LucasCioffi: Kevin mentioned we can keep it since we are standing up a new database in Production.  If Production is going to run the migrations, then `test_column` won't show up.  If Production is going to load the schema, then we do.  If the `test_column` data isn't important, then yes, it's a good ideas to create a PR with a migration to remove it.

app/controllers/registrations_controller.rb
```diff
@@ -11,8 +11,8 @@ def create
 		Rails.logger.debug("creating new user")
 		build_resource(sign_up_params)
 		if resource.valid?
-			begin
-				resource.send_request_to_patron_creator_service
+  		begin
+				resource.send_user_to_ils
```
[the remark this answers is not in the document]
nnat425: I agree.

test/unit/user_test.rb
```diff
@@ -4,7 +4,9 @@ class UserTest < ActiveSupport::TestCase
   include AwsDecrypt
 
   setup do 
-   @user = crank(:user)
+   @school = crank!(:school)
+   crank!(:user, :school_id => @school.id)
```
LucasCioffi: I think you can remove this line.

test/integration/user_flow_test.rb
```diff
@@ -1,25 +1,41 @@
...
+   @user = crank!(:user, :school_id => @school.id)
+end
 
-  email = generate_email
-
-  school = crank!(:school)
-  user = crank!(:user)
+  # Test that generated barcode is a 14-digit string, starting with '27777'.
+  test "user method assign_barcode increments last valid barcode by 1" do
+    user_two = crank!(:user, :school_id => @school.id)
```
LucasCioffi: More modern Ruby format (after 1.9): `user_two = crank!(:user, school_id: @school.id)`

test/integration/user_flow_test.rb
```diff
@@ -1,25 +1,41 @@
 require 'test_helper'
 
 class UserFlowTest < ActionDispatch::IntegrationTest
+ 
+setup do 
+   @school = crank!(:school)
+   @user = crank!(:user, :school_id => @school.id)
```
LucasCioffi: More modern Ruby format (after 1.9): `@user = crank!(:user, :school_id => @school.id)`

test/unit/user_test.rb
nnat425: We cannot. Reason being, if the test database is empty, the method called assign_barcode will throw an error since there is a query in that method to return a user record from the database.

PR #13, opened 2018-08-06
Edit breadcrumb when users edits their account MLN-146

LucasCioffi:
The breadcrumbs now say "Account", similar to when you go to /account
screen shot 2018-08-06 at 4 37 54 pm

### DIFF
--- a/app/views/layouts/_breadcrumb.html.erb
+++ b/app/views/layouts/_breadcrumb.html.erb
@@ -1,7 +1,7 @@
 <div id="breadcrumb">
 	<%= link_to "Home", "/app#/teacher_sets" %> <i class="fa fa-angle-right"></i>
 	<% case request.fullpath %>
-	<% when "/account" %>
+	<% when "/account", "/users/edit" %>
 	Account
 	<% when "/help" %>
 	Help


nodanaonlyzuul:
Hallo! Thank you for inviting me to the pull request party.
I'm weary of big conditional operations in a view/template...for now, this is a 2-case case statement.
It's totally fine for now but...

## If one day...

* You find yourself adding many more `when`s.
* The `href`'s value starts changing conditionally.
* This needs to be used on multiple pages (like in the layout erb)

I could see this becoming a helper (easily unit-testable!) like:

```ruby
def bread_crumbs(current_controller)
  currently_on = "#{current_controller.controller_name}##{current_controller.action_name}"
  mapping = {
    'sessions#new' => {href: users_path, link_text: "All Users"},
    'users#index' =>  {href: blogs_path, link_text: "All Blogs"},
  }

  link_to mapping[currently_on][:link_text], mapping[currently_on][:href]
end
```

And called like:

```erb
<%= bread_crumbs(controller) %>
```

LucasCioffi:
Thank you @nodanaonlyzuul this is very helpful!

PR #23, opened 2018-08-10
add accessibility fixes for screen readers

nnat425:
- This PR allows voice-over functionality when navigating the form.


app/views/layouts/_navigation.html.erb
```diff
@@ -4,7 +4,7 @@
     <% if user_signed_in? %>
     <%= link_to "<span>Dashboard</span>".html_safe, app_url, :class=>"link-home" %>
```
LucasCioffi: Looks like you can add `, "aria-label" => "Home-Logo"` to this line too.

The relevant CSS shows they use the same logo as the background image:
```
&.link-home {
      float: left;
      margin-left: 5px;
      width: 130px;
      background: url(/assets/mylibrarynyc-logo-white-mini.png) center no-repeat;
```

nnat425: Adding it.

PR #27, opened 2018-08-10
Fix duplicate email validation message and breadcrumb

LucasCioffi:
# Before
screenshot 2

# After
screen shot 2018-08-10 at 1 32 44 pm


LucasCioffi:
@nnat425 apparently Devise's validation is redundant with our own uniqueness validation on email according to comments [here](https://github.com/indentlabs/notebook/issues/235), so I removed it.  I tried to create an account with an email that is already taken, and validation is still working.  Do you get the same results?

nnat425:
@LucasCioffi I got the same results. I'd suggest adding some documentation in the user.rb file, letting other developers know that :validatable is checking e-mail uniqueness, so they are aware.

PR #28, opened 2018-08-10
Send pcodes 1-3 to patron creator service

LucasCioffi:


app/models/user.rb
```diff
@@ -107,8 +107,12 @@ def send_request_to_patron_creator_service
     query = {
       'names' => [last_name.upcase + ', ' + first_name.upcase],
       'emails' => [email],
+      'pin' => pin,
       'patronType' => patron_type,
       'patronCodes' => {
+        'pcode1' => '-',
+        'pcode2' => '-',
+        'pcode3' => pcode3,
```
nnat425: We should have consistency in regards to the the method names. We have patron_type, pcode3 and assign_barcode. I'd suggest starting everything with assign_keyName. Reason for that is, one can assume that the value of pcode3 and patronType are being generated from parameters of the form, which they aren't.

LucasCioffi: (We reached consensus in-person to leave as-is so that `pcode3` and `patron_type` can serve as getter methods going forward).

PR #31, opened 2018-08-13
Fixing broken pages in admin dashboard MLN-143

LucasCioffi:
I removed references to campus_id in the admin dashboard, because we are not importing that data for new schools anymore.  Courtney confirmed this is OK.

I replaced `.size` with `.count` because it was preventing the admin holds show page from loading.


app/models/school.rb
```diff
@@ -1,8 +1,4 @@
 class School < ActiveRecord::Base
-
-  attr_accessible :name
-
-  belongs_to :campus
```
nnat425: Hmm I'm a bit concerned we're removing this association. Are we 100% sure the association is not being called anywhere? In addition, why was campus initially added? What value did it provide besides for that fact of illustrating some schools that are located within the same area or build.

LucasCioffi: Yes, I think it's safe to remove campus_id because of Courtney's comments: https://jira.nypl.org/browse/MLN-143

When you get a chance, let me show you in-person what I found when I searched the source code for dependencies on campus_id.

PR #34, opened 2018-08-13
Fix alt email bug

nnat425:
- Add alt_email uniqueness back.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,7 +20,7 @@ class User < ActiveRecord::Base
   # created on sign up. Does not occur when updating
   # the record.
   validates_format_of :email, with: /\@schools.nyc\.gov/, message: ' should end in @schools.nyc.gov', :on => :create
-  validates_format_of :first_name, :last_name, :with => /^[a-z]+$/i
+  validates_format_of :first_name, :last_name, :with => /\A[^0-9`!@#\$%\^&*+_=]+\z/
   validates_format_of :alt_email,:with => Devise::email_regexp, :allow_blank => true, :allow_nil => true
   # Validation only occurs when a user record is being
   # created on sign up. Does not occur when updating
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,7 +15,7 @@ class User < ActiveRecord::Base
   attr_accessor :pin
 
   validates :school, :first_name, :last_name, :presence => true
-  validates :email, :uniqueness => true
+  validates :email, :alt_email, :uniqueness => true
   # Validation only occurs when a user record is being
   # created on sign up. Does not occur when updating
   # the record.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,17 +14,14 @@ class User < ActiveRecord::Base
   # Makes getters and setters
   attr_accessor :pin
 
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -14,17 +14,14 @@ class User < ActiveRecord::Base
   # Makes getters and setters
   attr_accessor :pin
 
-  validates :school, :first_name, :last_name, :presence => true
-  validates :email, :uniqueness => true
-  # Validation only occurs when a user record is being
+  # Validation's for email and pin only occurs when a user record is being
   # created on sign up. Does not occur when updating
   # the record.
+  validates :school, :first_name, :last_name, :presence => true
```
LucasCioffi: We can change this to `presence: true`

app/models/user.rb
```diff
@@ -14,17 +14,14 @@ class User < ActiveRecord::Base
...
   validates_format_of :email, with: /\@schools.nyc\.gov/, message: ' should end in @schools.nyc.gov', :on => :create
-  validates_format_of :first_name, :last_name, :with => /^[a-z]+$/i
+  validates_format_of :first_name, :last_name, :with => /\A[^0-9`!@#\$%\^&*+_=]+\z/
   validates_format_of :alt_email,:with => Devise::email_regexp, :allow_blank => true, :allow_nil => true
-  # Validation only occurs when a user record is being
-  # created on sign up. Does not occur when updating
-  # the record.
+  validates :alt_email, :uniqueness => true, :allow_blank => true, :allow_nil => true
```
LucasCioffi: We can change this to `validates :alt_email, uniqueness: true, allow_blank: true, allow_nil: true`

LucasCioffi: Please add tests for validation.

app/models/user.rb
```diff
@@ -14,17 +14,14 @@ class User < ActiveRecord::Base
...
   # the record.
+  validates :school, :first_name, :last_name, presence: true
   validates_format_of :email, with: /\@schools.nyc\.gov/, message: ' should end in @schools.nyc.gov', :on => :create
-  validates_format_of :first_name, :last_name, :with => /^[a-z]+$/i
+  validates_format_of :first_name, :last_name, :with => /\A[^0-9`!@#\$%\^&*+_=]+\z/
   validates_format_of :alt_email,:with => Devise::email_regexp, :allow_blank => true, :allow_nil => true
-  # Validation only occurs when a user record is being
-  # created on sign up. Does not occur when updating
-  # the record.
+  validates :alt_email, uniqueness: true, allow_blank: true, allow_nil: true
```
LucasCioffi: Looks good.  Let's remember to tell the product team that alt_email uniqueness doesn't take emails into account, just alt_emails.  We can create a custom validation if it is necessary.

daryanypl:
Hey folks, 
I'm not sure we should be restricting alt_email to be unique.  A teacher could have a personal account with that alt_email address.

LucasCioffi:
@daryanypl are you commenting from the plane?!

nnat425:
@daryanypl That's fine. alt_email is only comparing it's value to all the other records that have alt_email. alt_email DOES NOT compare it's value to every record containing e-mail.

PR #35, opened 2018-08-14
Toggle school's active status from admin dashboard MLN-167

LucasCioffi:
Admin user can now toggle active status on/off for schools in the admin dashboard.

screen shot 2018-08-14 at 9 17 55 am


app/models/school.rb
```diff
@@ -10,4 +11,21 @@ def full_name(delim = ', ')
     n += "#{delim}#{borough}" if !borough.nil?
     n
   end
+
+  def activation_links_container(async_boolean = false)
```
nnat425: We shouldn't be generating html through the model. I'd suggest creating a partial view and embedding this information there.

LucasCioffi:
Replaced by https://github.com/NYPL/MyLibraryNYCApp/pull/35

PR #37, opened 2018-08-15
Create table to map sierra codes and zcodes MLN-123

LucasCioffi:
This rake task will overwrite all joins between sierra_codes and zcodes:
`rake ingest:overwrite_sierra_code_zcode_joins['data/public/sierra_code_zcode_joins.csv']`
 
Antonio has given me this [spreadsheet](https://docs.google.com/spreadsheets/d/1ZYNRAFp1PESuELv778FXwC6U2CStmAfwsLFXM0-XMxo/edit?ts=5b6b180a#gid=0) which he exported from Sierra.  I removed all rows that were missing a zcode; they were just placeholders where zcodes can be added later.

We are asking him to add more joins in Sierra between `pcode4` and `sierra_codes`, so we will eventually run the rake task again to overwrite our join table.  Nothing depends on the ids in the join table, just the values of `sierra_code` and `zcode`, so this is OK.


db/migrate/20180810033347_create_sierra_code_zcode_joins.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSierraCodeZcodeJoins < ActiveRecord::Migration
```
nnat425: The name of migration and class are mis-leading to me. This is technically not a join table if both values are being derived from the same table without a join.

lib/tasks/ingest.rake
```diff
@@ -423,7 +423,7 @@ namespace :ingest do
         school.updated_at = Time.zone.now
         school.save
 
-        break if Rails.env.test? && index == 2 # import_schools_test.rb assumes that the CSV only has one record
+        break if Rails.env.test? # import_schools_test.rb only needs one record
```
nnat425: Why did you let import_schools_test.rb test one record vs opposed to all of them?

LucasCioffi: My goal was to test that the rake task saves a record properly rather than the quality of the data.  Testing one can save some time.  Sound reasonable?

db/migrate/20180810033347_create_sierra_code_zcode_joins.rb
LucasCioffi: Sounds good, @nnat425, I made the change from "joins" to "matches" as we described.

LucasCioffi:
@nnat425 FYI I added two new CSVs that are specific to the tests.

db/schema.rb
```diff
@@ -131,6 +130,13 @@
   add_index "schools", ["campus_id"], :name => "index_schools_on_campus_id"
   add_index "schools", ["code"], :name => "index_schools_on_code", :unique => true
 
+  create_table "sierra_code_zcode_matches", :force => true do |t|
+    t.string   "sierra_code"
```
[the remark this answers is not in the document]
LucasCioffi: The timestamps have database-level validations by default.  We'd have to do some extra work to add others.  It's a good idea if we have time later.

LucasCioffi:
@daryanypl ready for another look.

LucasCioffi:
Replaced by https://github.com/NYPL/MyLibraryNYCApp/pull/40.

LucasCioffi:
@daryanypl any unaddressed comments on this one?

LucasCioffi:
Hi @daryanypl I addressed the comments.

LucasCioffi:
Replaced by https://github.com/NYPL/MyLibraryNYCApp/pull/47

PR #38, opened 2018-08-15
add validation for pin

nnat425:
- Add validation for PIN format 
- Patterns such as:  

- 1111
- 9191
- 0008
- 7000

are prohibited.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,7 +107,7 @@ class User < ActiveRecord::Base
     if pin.scan(/(.)\1{2,}/).empty? && pin.scan(/(..)\1{1,}/).empty? == true
       true
     else
-      errors.add(:pin, 'format is invalid.')
+      errors.add(:pin, 'does not meet our requirements. Please try again.')
       false
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -96,6 +97,20 @@ def assign_barcode
...
+  # the following examples:
+  # 1111, 2929, 0003, 5999. 
+  # Sierra will return the following
+  # error message if PIN is invalid:
+  # "PIN is not valid : PIN is trivial"
+  def validate_pin_pattern
+    if pin.scan(/(.)\1{2,}/).empty? && pin.scan(/(..)\1{1,}/).empty? == true
+      true
+    else
+      errors.add(:pin, 'format is invalid.')
```
daryanypl: How about this message:  "Pin must be a 4-digit number with no repeating segments."

nnat425: Let me run it by Courtney.

nnat425: Talked to Courtney. We're going with "Your PIN does not meet our requirements. Please try again." .

PR #41, opened 2018-08-16
Execute all tests

nnat425:
- Have command 'rake tests:all' run both unit and integration tests.


nnat425:
- Currently have another PR with executing tests.

PR #42, opened 2018-08-16
Set up prod

nnat425:
- Make my-library-nyc-application.log the default log output for all environments 
- Add prod credentials


config/environments/production.rb
```diff
@@ -28,7 +28,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = false
```
daryanypl: Not sure about this?

nnat425: @daryanypl  This has been addressed via your conversation with Kevin. Just noting that here.

PR #43, opened 2018-08-16
Latest changes from dev to qa

nnat425:


app/admin/schools.rb
```diff
@@ -1,28 +1,51 @@
-ActiveAdmin.register School do
+ActiveAdmin.register School do |aa|
```
daryanypl: we don't use aa in the block, though, right?

daryanypl: does ActiveAdmin.register School ever fail?  what happens if it does?

app/models/school.rb
```diff
@@ -11,15 +7,7 @@ class School < ActiveRecord::Base
...
-    if !campus.nil?
-      if campus.schools.size > 1
-        n += "#{delim}#{campus.name}"
-      end
-
-      if !campus.borough.nil?
-        n += "#{delim}#{campus.borough.name}"
-      end
-    end
+    n += "#{delim}#{borough}" if !borough.nil?
```
daryanypl: Are campuses being removed from the model?

app/models/school.rb
```diff
@@ -11,15 +7,7 @@ class School < ActiveRecord::Base
 
   def full_name(delim = ', ')
```
daryanypl: Needs method doc.

app/models/user.rb
```diff
@@ -14,18 +14,17 @@ class User < ActiveRecord::Base
...
   # the record.
+  validates :school_id, :first_name, :last_name, :presence => true
   validates_format_of :email, with: /\@schools.nyc\.gov/, message: ' should end in @schools.nyc.gov', :on => :create
-  validates_format_of :first_name, :last_name, :with => /^[a-z]+$/i
+  validates_format_of :first_name, :last_name, :with => /\A[^0-9`!@;#\$%\^&*+_=\x00-\x19]+\z/
   validates_format_of :alt_email,:with => Devise::email_regexp, :allow_blank => true, :allow_nil => true
-  # Validation only occurs when a user record is being
-  # created on sign up. Does not occur when updating
-  # the record.
+  validates :alt_email, uniqueness: true, allow_blank: true, allow_nil: true
```
daryanypl: Is the doc string deleted because it;s no longer true?  If so, could we modify the doc string, rather than delete?

test/unit/user_test.rb
```diff
@@ -12,31 +12,64 @@ class UserTest < ActiveSupport::TestCase
...
+      @user.save
+      assert_equal(@user.errors.messages[:pin],["does not meet our requirements. Please try again."])
+    end
+  end
+
+  [generate_email].each do |new_email|
+    test 'user model cannot be created with 3 of the same repeated digits in a row as pin' do
+      @user.pin = "0007"
+      @user.save
+      assert_equal(@user.errors.messages[:pin],["does not meet our requirements. Please try again."])
```
daryanypl: But I think we changed the error msg wording, right?

nnat425: We changed it this per Courtney's approval.

app/models/user.rb
nnat425: The doc string was being presented twice. I edited line 17 to be specific. It's cleaner this way.

app/models/school.rb
LucasCioffi: @daryanypl we're leaving campus data alone just in case we need it in the future.  We're not importing campus info for new schools, though.

app/admin/schools.rb
LucasCioffi: @daryanypl I removed `|aa|` in https://github.com/NYPL/MyLibraryNYCApp/pull/59

I think that it safe to assume that `ActiveAdmin.register` will always succeed.  It's one of thousands of small steps in setting up the app's environment and it's convention to assume they all will pass.  I think it would take something like a typo to prevent successful initialization, and we'd expect the app's initialization to fail.  So I think we're good here.

PR #46, opened 2018-08-20
Speed up travis

nnat425:
- Speeds up build time for a travis-ci deployment.


nnat425:
Information about travis caching can be found here: https://docs.travis-ci.com/user/caching/

.travis.yml
```diff
@@ -1,11 +1,11 @@
-language: generic
-cache: bundler
+language: ruby
```
LucasCioffi: @nnat425 do you remember why we used "generic" here before?  I thought it was essential for Travis to work, but I don't remember the specifics.

holingpoon: @LucasCioffi, `language: generic` was initially set up by @kfriedman. Originally, travis returns failures when `language: ruby` was set.

nnat425: Just so this is noted: In order for the caching to work, I learned that the language needed to be switched to ruby. Otherwise, it would throw me errors on generic.

PR #49, opened 2018-08-20
fix bugs

nnat425:
- Text for no javascript error message is now not uppercase 
- Fix async check issue when e-mail is already verified 
- Add some logging

### DIFF
--- a/app/assets/javascripts/sign_up_async_request.js.erb
+++ b/app/assets/javascripts/sign_up_async_request.js.erb
@@ -5,10 +5,17 @@ $( document ).ready(function() {
 
 	$('#user_email').on('focusout', function(){
 
+		if(localStorage.getItem("valid-email") == $('#user_email').val()){
+			$("#check_email_div").css("display", "block")
+			validEmailMessage();
+			return true
+		}
+
 		if(ajaxRequestArray.length > 0){
 			ajaxRequestArray[0].abort();
 			ajaxRequestArray.pop();
 		}
+
 
 		$("#sign_up_button").prop("disabled", true);
 		$("#sign_up_button").css("background-color", "grey");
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/sign_up_async_request.js.erb
```diff
@@ -78,6 +78,21 @@ function handleError(){
 	$("#check_email_span").text("We've encountered an error. Please try again later or email help@mylibrarynyc.org for assistance.");
 }
 
+function validEmailMessage(){
```
LucasCioffi: `showValidEmailMessage` might be a bit more accurate for what it does.

LucasCioffi: Maybe pass in a boolean value about whether the validation passed or failed, and then you can fully control the disabling of the form's submit button here.

app/models/user.rb
```diff
@@ -123,6 +123,12 @@ def send_request_to_patron_creator_service
         }
       ],
     }
+    Rails.logger.debug(
+       {
+        'status' => 'Request send to patron creator service',
+        'dataSent' => query
```
daryanypl: I'd rather not send the query to the log without passing it through a safety filter first, but we can fix this after we erplace the logging with our wrapper.

app/views/devise/registrations/new.html.erb
```diff
@@ -53,7 +52,7 @@
   </div>
 
   <div class="field-group">
-    <%= label_tag :user_school_id, 'Your school:', :id => 'userSchool-label' %>
+    <%= label_tag :user_school_id, 'Your school', :id => 'userSchool-label' %>
```
daryanypl: Using a label for id info might be problematic for accessibility, but let's write this down and move forward and do accessibility when we have Willa.

app/models/user.rb
nnat425: I agree.

PR #57, opened 2018-08-23
Rename migrations so they will run after QA & Dev DB refresh MLN-214

LucasCioffi:
We're re-populating the database in the QA and Dev environments so that we can run the `import_all_nyc_schools` rake task and delete extra schools via migrations.  This will simulate what will happen in Production and help us catch any errors ahead of time.


LucasCioffi:
No longer needed.

PR #62, opened 2018-08-24
Log on email send

daryanypl:


app/mailers/hold_mailer.rb
```diff
@@ -1,29 +1,55 @@
...
+      @teacher_set = hold.teacher_set
+
+      emails = AdminUser.pluck(:email)
+      Rails.logger.debug("#{LOG_TAG}.admin_notification: About to send hold order notification email on #{@teacher_set.title or 'unknown'} to #{@user.email or 'unknown'}")
+      mail(:to => emails, :subject => "New Order from #{@user.email or 'unknown'} for #{@teacher_set.title or 'unknown'}")
+    rescue => exception
+      # something went wrong.  perhaps either user or teacher set aren't filled properly.
+      # or maybe the email couldn't be sent out.
+      Rails.logger.error("#{LOG_TAG}.admin_notification: Cannot send hold order notification email.  Backtrace=#{exception.backtrace}.")
+      raise
```
LucasCioffi: @daryanypl I think it's preferable to let the user's hold get placed without raising an error for them, because we will have an alarm and we can take action to fix the problem.

If the user sees the error, it's more inconvenient for them, and they may not come back to order their books.  This might be something to bring up in a discussion with Courtney or Kevin.

daryanypl: Agreed, we should talk about it.

PR #65, opened 2018-08-24
Merging latest (logging and JS version-specific changes) from development to qa

daryanypl:


LucasCioffi:
@daryanypl looks like all these changes were approved and merged into the development branch already, so you can go ahead and merge to QA without our approval.  

@nnat425 and I make the title of PRs like this something like "Merging latest from Dev to QA" and then we merge it in automatically.

PR #67, opened 2018-08-24
Adding SSL certificate to MLN load balancer from ebextensions file

daryanypl:
This change is already merged to master, but merging it properly into development here.

### DIFF
--- a/.ebextensions/07_https-nypl-digital-dev.config
+++ b/.ebextensions/07_https-nypl-digital-dev.config
@@ -1,0 +1,4 @@
+option_settings:
+  aws:elb:loadbalancer:
+    LoadBalancerHTTPSPort: 443
+    SSLCertificateId: arn:aws:acm:us-east-1:946183545209:certificate/d38c5c82-708d-4942-a4cc-bc4985a9a4c7


.ebextensions/07_https-nypl-digital-dev.config
```diff
@@ -0,0 +1,4 @@
+option_settings:
+  aws:elb:loadbalancer:
+    LoadBalancerHTTPSPort: 443
+    SSLCertificateId: arn:aws:acm:us-east-1:946183545209:certificate/d38c5c82-708d-4942-a4cc-bc4985a9a4c7
```
kfriedman: Hmm... you're merging to the `development` branch but your SSL ID refers to the production environment (`946183545209`). 

Which brings up a point: should we also create a certificate in the development (`nypl-sandbox`) environment?

daryanypl: Yeah, I'm expecting this code to fail, and then I think I'll deactivate it until we can get the stage-sensitive deployment options.

daryanypl: I'd like to leave a status report here.  

I pushed this file to development.  I expected development to fail.  Development did not fail.  This is a mystery, but I'll take it.  Looks like the ebextensions file is ignored when there's no load balancer to set the ssl up on.

I'm keeping this file in the code tree, even though it's not actually doing anything on either development or production.

Related note:  Pushing log retention settings in ebextensions also doesn't overwrite the log retention settings made manually in CloudWatch console.

PR #76, opened 2018-08-28
Merging corrected platform oauth creds from development to qa

daryanypl:


config/environments/production.rb
```diff
@@ -15,7 +15,7 @@
   config.assets.compress = true
 
   # Don't fallback to assets pipeline if a precompiled asset is missed
-  config.assets.compile = false
+  config.assets.compile = true
```
nnat425: This is already set to true in the qa environment and production. Why is it github saying otherwise?

PR #83, opened 2018-08-28
Merging sendgrid host env variable fix from qa to master

daryanypl:


config/environments/qa.rb
```diff
@@ -0,0 +1,7 @@
+MyLibraryNYC::Application.configure do
+  # host doesn't matter, it only matters that it exists (for dev and qa, for production it does matter)
+  config.action_mailer.default_url_options = { :host => 'localhost:3000' }
+  config.action_mailer.perform_deliveries = true
+
+  Rails.env = 'qa'
```
LucasCioffi: I know that AWS EB is setting RAILS_ENV and RACK_ENV (which I found overrides RAILS_ENV through last Friday's testing), so is `Rails.env = 'qa'` necessary?

nnat425: I agree, most likely not. But more importantly, post deployment we should make our qa.rb file  very similar to our production.rb file.

PR #88, opened 2018-08-29
extend time of flash message if error

nnat425:
- When the micro-service or oauth token service returns an invalid response to MLN, we are now rendering an existing flash message for 15 seconds


app/views/layouts/_messaging.html.erb
```diff
@@ -1,5 +1,9 @@
 <% flash.each_with_index do |(name, msg), i| %>
+ <% if name == :registration_error %>
+  <div id="alert-<%=i%>" data-alert class="alert main-alert fade-in-out-longer <%= name %>">
+ <% else %>
   <div id="alert-<%=i%>" data-alert class="alert main-alert fade-in-out <%= name %>">
-    <%= msg %>
+    <% end %>
+       <%= msg %>
   </div>
```
LucasCioffi: Suggestion:
```
<% flash.each_with_index do |(name, msg), i| %>
  <div id="alert-<%=i%>" data-alert class="alert main-alert <%= (name == :registration_error ? 'fade-in-out-longer' : 'fade-in-out') %> <%= name %>">
    <%= msg %>
  </div>
<% end %>
```

PR #92, opened 2018-08-29
Changed the ssl certificate to the wildcard one.

daryanypl:


daryanypl:
Seems so.  If the ssl certificate ebextensions files weren't ignored, we'd have a broken build on dev, where the certificate arn is invalid.  It appears that, once you set up ssl in the aws console, the ebextensions is ignored.  But still good to have for documentation in case environment needs to be re-created.

PR #96, opened 2018-08-31
Send hold cancellation email even if no reason MLN-236

LucasCioffi:
Previously, if the user didn't enter a reason for cancellation, they didn't receive an email.  We want the email to be sent in both situations.

I also added local.rb because emails weren't being sent without this being set: `config.action_mailer.default_url_options = { :host => 'localhost:3000' }` and without that file, it wasn't set.

I also turned off `send_admin_notification_email` if the user is in a local or development environment.

# Order & Cancellation Emails
screen shot 2018-08-31 at 3 20 02 pm

# Cancellation Email
screen shot 2018-08-31 at 3 20 11 pm

### DIFF
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -81,6 +81,7 @@ class Hold < ActiveRecord::Base
   end
 
   def send_admin_notification_email
+    return if Rails.env.development? || Rails.env.local?
     HoldMailer.admin_notification(self).deliver
   end
 
--- a/app/models/hold_change.rb
+++ b/app/models/hold_change.rb
@@ -8,15 +8,15 @@ class HoldChange < ActiveRecord::Base
   belongs_to :admin_user
 
   after_save :do_after_save
-  
+
   def do_after_save
     update_hold
     send_change_status_email
   end
-  
+
   def send_change_status_email
-    # deliver email if there's a comment and status has been changed to error, pending, or closed
-    HoldMailer.status_change(hold, status, comment).deliver if !comment.blank? and ['error', 'pending', 'closed', 'cancelled'].include? status
+    # deliver email if status has been changed to error, pending, closed, or cancelled
+    HoldMailer.status_change(hold, status, comment).deliver if ['error', 'pending', 'closed', 'cancelled'].include? status
   end
 
   def update_hold
@@ -24,5 +24,5 @@ class HoldChange < ActiveRecord::Base
     hold.status = status
     hold.save
   end
-  
+
 end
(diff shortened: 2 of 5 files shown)


app/models/hold.rb
```diff
@@ -81,6 +81,7 @@ def order_number
   end
 
   def send_admin_notification_email
+    return if Rails.env.development? || Rails.env.local?
```
nnat425: What is your thought process for not sending the e-mail in development and local? I'd preferably like to have a separate e-mail account for all environments. Or the same e-mail account for local/dev/qa. I believe production shouldn't be used in the lower environments.

LucasCioffi: Yes makes sense to me.  @daryanypl may I pass this off to you so that you can decide what is best (not sending emails or creating new accounts to receive messages from different environments)?

daryanypl: I like the idea of having separate email addresses to test with.  Lets accept this code as the immediate fix, and then add a Jira ticket to the backlog (I'll write it).

app/models/hold_change.rb
```diff
@@ -16,7 +16,7 @@ def do_after_save
   
   def send_change_status_email
     # deliver email if there's a comment and status has been changed to error, pending, or closed
```
daryanypl: Need to change the doc string to reflect the code change.

app/views/hold_mailer/status_change.html.erb
```diff
@@ -1,4 +1,4 @@
-<p><%= @message %></p>
+Your order status for <%= @teacher_set.title %> is <%= @hold.status %>.
```
daryanypl: Is the message var no longer used for anything?

config/environments/local.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  # Log error messages when you accidentally call methods on nil.
+  config.whiny_nils = true
+
+  # Show full error reports and disable caching
+  config.consider_all_requests_local       = true
+  config.action_controller.perform_caching = false
+
+  # Don't care if the mailer can't send
+  config.action_mailer.raise_delivery_errors = true
```
daryanypl: Does the doc string correspond to the true value?

LucasCioffi: I'll remove it to be safe.  It's ambiguous.

LucasCioffi:
@daryanypl I checked w/ Courtney and she's OK with adding the message/comment by the user back in.  You're right in that if I remove it from the view, I should have removed it from the code which prepares the mailer. New screenshot:
screen shot 2018-09-04 at 3 51 54 pm

LucasCioffi:
@daryanypl @nnat425 any other thoughts prior to approval?

PR #104, opened 2018-09-06
fix issue of text jump

nnat425:
- Task references ticket https://jira.nypl.org/browse/MLN-67 
- Fixes jumping in text


LucasCioffi:
@nnat425 before & after screenshots would be helpful for these visual fixes.

PR #111, opened 2018-09-06
Change order of when emails are sent

nnat425:


app/mailers/hold_mailer.rb
```diff
@@ -9,6 +9,7 @@ def admin_notification(hold)
     begin
       @hold = hold
       @user = hold.user
+      @break = '123' + 123
```
nnat425: This will be removed once Chris verifies on QA.

PR #121, opened 2018-09-10
add ses config

nnat425:


app/controllers/holds_controller.rb
```diff
@@ -52,7 +52,7 @@ def cancel
   end
 
   def create
-    begin
+      begin
```
daryanypl: fix whitespace

config/environments/development.rb
```diff
@@ -36,7 +35,6 @@
   config.assets.debug = true
 
   config.action_mailer.default_url_options = { :host => 'my-library-nyc-app-development.us-east-1.elasticbeanstalk.com' }
-  config.action_mailer.perform_deliveries = true
```
daryanypl: Is config.action_mailer.perform_deliveries no longer a setting?

config/environments/production.rb
```diff
@@ -73,10 +73,10 @@
...
   config.action_mailer.smtp_settings = {
    :port =>           '587',
-   :address =>        'smtp.sendgrid.net',
-   :user_name =>      ENV['SENDGRID_USERNAME'],
-   :password =>       ENV['SENDGRID_PASSWORD'],
-   :domain =>         'heroku.com',
+   :address =>        'email-smtp.us-east-1.amazonaws.com',
+   :user_name =>      ENV['SMTP_USERNAME'],
+   :password =>       ENV['SMTP_PASSWORD'],
+   :domain =>         'mylibrarynyc.org',
```
daryanypl: Is there email setting for local/dev/qa?

config/initializers/setup_mail.rb
```diff
@@ -1,10 +1,11 @@
...
+ActionMailer::Base.delivery_method = :smtp	
+ActionMailer::Base.perform_deliveries = true	
+
+ActionMailer::Base.smtp_settings = {	
+  address:        'email-smtp.us-east-1.amazonaws.com',
+  port:           '587',	
+  authentication: :plain,	
+  user_name:      ENV['SMTP_USERNAME'],	
+  password:       ENV['SMTP_PASSWORD'],	
+  domain:         'mylibrarynyc.org',
```
daryanypl: This block seems to conflict with production.rb?

config/environments/development.rb
nnat425: No. It's being set config/initializers/setup_mail.rb

config/environments/production.rb
nnat425: Technically, all environments email settings are being called in  config/initializers/setup_mail.rb . I can remove this code.

config/initializers/setup_mail.rb
nnat425: Keeping this block. And removing the block in the production.rb file.

config/initializers/setup_mail.rb
```diff
@@ -1,10 +1,11 @@
...
-ActionMailer::Base.smtp_settings = {
-  address:        'smtp.sendgrid.net',
-  port:           '587',
-  authentication: :plain,
-  user_name:      ENV['SENDGRID_USERNAME'],
-  password:       ENV['SENDGRID_PASSWORD'],
-  domain:         'heroku.com',
-  enable_starttls_auto: true
+ActionMailer::Base.delivery_method = :smtp	
+ActionMailer::Base.perform_deliveries = true
```
LucasCioffi: @nnat425 maybe we shouldn't perform deliveries in local and development, because we try a lot of test accounts like test@example.com and we wouldn't want to send emails there.

config/environments/production.rb
```diff
@@ -65,19 +65,7 @@
   # with SQLite, MySQL, and PostgreSQL)
   # config.active_record.auto_explain_threshold_in_seconds = 0.5
 
+  config.action_mailer.perform_deliveries = false
```
LucasCioffi: Did you mean to add this line into local.rb instead?
I would think we'd want this to be `true` here.

PR #122, opened 2018-09-11
Ability to add user via MLN admin interface MLN-279

LucasCioffi:
There is a page for adding a new user via the admin dashboard:
screen shot 2018-09-11 at 9 02 22 am

The required fields are the ones that are the same from the sign-up-form:
screen shot 2018-09-11 at 9 05 17 am

The system automatically assigns a barcode if none is assigned by the admin:
screen shot 2018-09-11 at 9 35 46 am

The admin can edit the user:
screen shot 2018-09-11 at 9 35 57 am

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -2,6 +2,10 @@ ActiveAdmin.register User do
   menu :priority => 6
   actions :all
 
+  before_create do |user|
+    user.send_request_to_patron_creator_service if user.valid? # if not valid, errors will show on the form
+  end
+
   index do
     column :email
     column :current_sign_in_at
@@ -22,27 +26,31 @@ ActiveAdmin.register User do
       f.input :first_name
       f.input :last_name
       f.input :email
-      f.input :barcode
       f.input :alt_email
-      f.input :alt_barcodes
+      if f.object.new_record?
+        f.input :barcode, label: raw('Barcode<br>(leave blank to auto-assign)')
+      else
+        f.input :barcode
+      end
+      f.input :pin, label: raw("4-digit PIN<br> It cannot contain a digit that is repeated 3 or more times (0001, 5555) and cannot be a pair of repeated digits (1212, 6363)")
     end
     f.actions
   end
 
-  show do |ad|
+  show title: proc{|user| user.name(true) } do |user|
 
     attributes_table do
       row :contact_email do
-        link_to ad.contact_email, ad.contact_email
+        link_to user.contact_email, user.contact_email
       end
       row :school do
-        if ad.school.nil?
+        if user.school.nil?
           'No school selected'
         else
-          link_to "#{ad.school.code}: #{ad.school.full_name(' / ')}", admin_school_path(ad.school)
+          link_to "#{user.school.code}: #{user.school.full_name(' / ')}", admin_school_path(user.school)
         end
       end
-      row :doe_email do ad.email end
+      row :doe_email do user.email end
       [:barcode, :alt_barcodes, :first_name, :last_name, :updated_at, :current_sign_in_at, :last_sign_in_at].each do |prop|
         row prop
       end
@@ -69,4 +77,9 @@ ActiveAdmin.register User do
 
   end
 
+  controller do
+    def edit # the edit page title has to be handled this way, source: https://github.com/activeadmin/activeadmin/wiki/Set-page-title
+      @page_title = resource.name(true)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -86,7 +86,7 @@ def assign_barcode
   # error message if PIN is invalid:
   # "PIN is not valid : PIN is trivial"
   def validate_pin_pattern
-    if pin.scan(/(.)\1{2,}/).empty? && pin.scan(/(..)\1{1,}/).empty? == true
+    if pin && pin.scan(/(.)\1{2,}/).empty? && pin.scan(/(..)\1{1,}/).empty? == true
```
daryanypl: Let's add the "non-trivial pin" validation, so admins don't input "1111" (see regex on Raj's user.rb").

app/admin/user.rb
```diff
@@ -22,27 +26,31 @@
       f.input :first_name
       f.input :last_name
       f.input :email
-      f.input :barcode
       f.input :alt_email
-      f.input :alt_barcodes
+      if f.object.new_record?
+        f.input :barcode, label: 'Barcode (leave blank to auto-assign)'
```
daryanypl: Nice feature.

app/admin/user.rb
```diff
@@ -2,6 +2,10 @@
   menu :priority => 6
   actions :all
 
+  before_create do |user|
+    user.send_request_to_patron_creator_service if user.school.present?
```
daryanypl: I'd like to see a code comment, explaining what this line is for, and what the correct path is, if school isn't present.

app/admin/user.rb
```diff
@@ -22,27 +26,31 @@
       f.input :first_name
       f.input :last_name
       f.input :email
-      f.input :barcode
       f.input :alt_email
-      f.input :alt_barcodes
+      if f.object.new_record?
```
daryanypl: I'm not sure what alt_barcodes does.  It feels like maybe it's redundant and should be deleted from both user model and admin user model.  If so, let's bring it back here, and make a new PR for that task.

LucasCioffi: @daryanypl Courtney was pleased to see alt_barcodes removed from the form.  There are 3000 users that have an alt_barcode values like this: `;277770XXXXXXXX;277770XXXXXXXX` or this: `277770XXXXXXXX` so maybe the best thing is just to leave the data in the database?

app/models/user.rb
LucasCioffi: @daryanypl it looks like the validation is already happening.
screen shot 2018-09-11 at 12 38 02 pm

app/admin/user.rb
daryanypl: If they're in the database and being used, we should provide a way to see and edit.  If they're in the database and being phased out, we should provide a way to see, but not edit (show the field, but gray it out and make it inactive with a little "deprecated" tag).  If we're getting rid of them, it should be its own separate pull request and user story.

The goal is to make sure that a pull request contains only the code that's relevant to that pull request.  None of us are perfect at it, but we try.

app/models/user.rb
LucasCioffi: Thanks @daryanypl here's what it looks like now:
screen shot 2018-09-12 at 2 30 49 pm

app/admin/user.rb
LucasCioffi: @daryanypl you mentioned verbally you'd create another ticket for this.  Thank you.
This is ready for your final review.

app/models/user.rb
```diff
@@ -112,7 +112,7 @@ def send_request_to_patron_creator_service
         'pcode3' => pcode3,
         'pcode4' => pcode4
       },
-      'barcodes' => [self.assign_barcode.to_s],
+      'barcodes' => [self.barcode.present? ? self.barcode : self.assign_barcode.to_s],
```
nnat425: Seems like your adding this logic to use this method on an existing user. Is that correct?

LucasCioffi: @nat425 if an admin enters a barcode for a patron when they are creating that patron through the admin dashboard, then we want to use that barcode in send_request_to_patron_creator_service, otherwise we want to assign a new barcode.

Previously, we were assigning barcodes within the `send_request_to_patron_creator_service` method for all patrons.

app/admin/user.rb
daryanypl: Yep, made https://jira.nypl.org/browse/MLN-308 .

PR #124, opened 2018-09-13
Synch Sierra Zcodes with MLN database MLN-288

LucasCioffi:


db/migrate/20180913135147_ingest_sierra_code_zcode_matches2.rb
```diff
@@ -0,0 +1,5 @@
+class IngestSierraCodeZcodeMatches2 < ActiveRecord::Migration
+  def up
+    Rake::Task['ingest:overwrite_sierra_code_zcode_matches'].invoke('data/public/sierra_code_zcode_matches.csv')
```
daryanypl: Am I correct in reading this as a destructive change?

LucasCioffi: @daryanypl Yes, when we overwrite the table, all the codes get new primary IDs, but that won't have an impact because we don't reference the primary IDs anywhere; we just send the code in the request to the Patron Creator Service.

data/public/sierra_code_zcode_matches.csv
```diff
@@ -1,972 +1,1880 @@
...
+zk430,444
+zk442,445
+zk490,446
+zk493,447
+zk502,449
+zk504,450
+zk507,451
+zk510,452
+zk590,455
+zm596,#N/A
```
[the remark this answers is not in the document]
LucasCioffi: @daryanypl Yes; great catch. I fixed it just now.

data/public/sierra_code_zcode_matches.csv
```diff
@@ -1,972 +1,1880 @@
...
+zx009,201
+zx391,202
+zk056,203
+zk122,204
+zk373,205
+zk445,206
+zk505,207
+zx207,208
+zx307,209
+Bridge Haven,210
```
daryanypl: Will ask Courtney about not having z code.

PR #126, opened 2018-09-14
fix logs

nnat425:


app/assets/stylesheets/vendor/jquery-ui.css
```diff
@@ -799,15 +799,15 @@ body .ui-tooltip {
 }
 .ui-widget-content {
 	border: 1px solid #aaaaaa;
-	background: #ffffff url(images/ui-bg_flat_75_ffffff_40x100.png) 50% 50% repeat-x;
+	background: #ffffff 50% 50% repeat-x;
```
LucasCioffi: @nnat425 does this have any change for the user?

app/models/user.rb
```diff
@@ -100,11 +113,10 @@ def validate_pin_pattern
   # a success/failure response.
   # Accepts a response from the microservice, and returns.
   def send_request_to_patron_creator_service
-    Rails.logger.debug("send_user_to_patron_creator_service: start")
     query = {
       'names' => [last_name.upcase + ', ' + first_name.upcase],
       'emails' => [email],
-      'pin' => pin,
+      'pin' => '1234',
```
LucasCioffi: Is this hardcoded change in here accidentally?

nnat425: Yes! Good catch.

app/assets/stylesheets/vendor/jquery-ui.css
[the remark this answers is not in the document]
nnat425: None at all. These images were not even stored in the repo from the start.

config/environments/local.rb
```diff
@@ -22,6 +22,9 @@
   # Only use best-standards-support built into browsers
   config.action_dispatch.best_standards_support = :builtin
 
+  # Turn off asset pipline information showing in logs 
+  config.assets.logger = false
```
nnat425: @daryanypl @LucasCioffi As mentioned per slack, setting this to true will turn off requested assets from being printed out in the logs. What are both of your thoughts on this? I'd like to get some perspective.

app/assets/stylesheets/vendor/jquery-ui.css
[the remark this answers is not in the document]
LucasCioffi: @nnat425 Ok then we can remove `50% 50% repeat-x` which positions the background image.  So all that is left is setting the color to `#ffffff`.

config/environments/local.rb
nnat425: Per conversation with Darya, we're going to leave it in for now.

PR #128, opened 2018-09-17
update csv file output

nnat425:
- Update CSV output per ticket - https://jira.nypl.org/browse/MLN-307

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -18,9 +18,9 @@ ActiveAdmin.register User do
     column :last_name
     column :alt_email
     column :alt_barcodes
-    column(:school_name) { |user| user.school.name}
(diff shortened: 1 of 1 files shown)


app/admin/user.rb
```diff
@@ -6,6 +6,23 @@
...
+    column :current_sign_in_at
+    column :last_sign_in_at
+    column :created_at
+    column :updated_at
+    column :barcode
+    column :first_name
+    column :last_name
+    column :alt_email
+    column :alt_barcodes
+    column(:school_name) { |user| user.school.name}
```
LucasCioffi: In case a user doesn't have a school for some reason, so these three lines could look like this:

`{ |user| user.school.present? ? user.school.name : '' }`

PR #133, opened 2018-09-19
Fix logger on dev

nnat425:


.travis.yml
```diff
@@ -26,7 +26,7 @@ deploy:
   bucket_path: my-library-nyc-app-development
   on:
     repo: NYPL/MyLibraryNYCApp
-    branch: development
+    branch: fix_logger_on_dev
```
LucasCioffi: Please remember to change this.

config/environments/local.rb
```diff
@@ -22,8 +22,8 @@
   # Only use best-standards-support built into browsers
   config.action_dispatch.best_standards_support = :builtin
 
-  # Turn off asset pipline information showing in logs 
-  config.assets.logger = false 
+  # # Turn off asset pipline information showing in logs 
+  # config.assets.logger = true
```
daryanypl: Let's bring this line back, and set it to false?

lib/wrapper.rb
```diff
@@ -0,0 +1,29 @@
+module Wrapper
```
daryanypl: Would you mind calling it LogWrapper, so it's more clear when you call it?

lib/wrapper.rb
```diff
@@ -0,0 +1,29 @@
...
+	# The method log take's in 2 parameters.
+	# A string passed in as level and a hash
+	# passed in as message. Please look at 
+	# examples from the user.rb model. 
+	
+	def self.log(level,message)
+
+	severity_levels = 
+	    {
+		'DEBUG' => 0,
```
daryanypl: Let's use the set from https://github.com/NYPL/engineering-general/blob/master/standards/logging.md .

lib/wrapper.rb
nnat425: Yes.

lib/wrapper.rb
nnat425: Look at line 25. We use a method called add. Add is a built in Ruby method. I'm not sure if it will work properly if you use the engineering-general standard numbers. Does that make sense? 

Read the first 2 paragraphs of the following link: https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels

config/environments/local.rb
nnat425: So I added this line originally. I can set it back to false but I believe it should be commented out. Are you ok with that?

PR #136, opened 2018-09-21
Make the timeout check IE compatible

LucasCioffi:
IE couldn't handle the XMLHttpRequest format, so I changed it to AJAX.

If you want to test it locally, use this in sessions_controller.rb:
```
  def max_session_duration
    20.seconds
  end

  def timeout_warning_duration
    10.seconds
  end
```
and this line twice in _timeout.html:
```
setTimeout(PeriodicalQuery, (3 * 1000));  //three seconds rather than 5 minutes
```

### DIFF
--- a/app/views/layouts/_timeout.html
+++ b/app/views/layouts/_timeout.html
@@ -1,19 +1,22 @@
 <script>
   function PeriodicalQuery() {
-    var request = new XMLHttpRequest();
-    request.onload = function (event) {
-      if (request.response['timeout_status'] == 'no_current_user') {
-        return;
-      } else if (request.response['timeout_status'] == 'timeout_warning') {
-        $('#timeout-warning').slideDown('fast');
-      } else if (request.response['timeout_status'] == 'timed_out') {
-        window.location.href = '/users/start';
-      }
-    };
-    request.open('GET', '/timeout_check', true);
-    request.responseType = 'json';
-    request.send();
+    $.ajax({
+      url: '/timeout_check',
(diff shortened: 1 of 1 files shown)


app/views/layouts/_timeout.html
```diff
@@ -1,19 +1,22 @@
...
-      } else if (request.response['timeout_status'] == 'timeout_warning') {
-        $('#timeout-warning').slideDown('fast');
-      } else if (request.response['timeout_status'] == 'timed_out') {
-        window.location.href = '/users/start';
-      }
-    };
-    request.open('GET', '/timeout_check', true);
-    request.responseType = 'json';
-    request.send();
+    $.ajax({
```
daryanypl: Sad performance, Internet Exploder.

Nice fix.  I agree with Navraj that I usually expect javascript in js files, but I think I remember from the last PR that you were including this html in an iframe?

LucasCioffi:
@nnat425 I looked into it after your comment.  To make it a JS partial, I'd have to remove the <script> tags but because it's included on an HTML page, it needs the <script> tags.  So keeping the script tags in makes it an HTML partial and enables it to be embedded into HTML, so I think we're OK justifying it as-is.

app/views/layouts/_timeout.html
LucasCioffi: This is included in two places: the angular layout and the default Rails layout.  It's included in an HTML page, so we can and should keep it in the HTML format.

PR #139, opened 2018-09-24
Rename books_in_sets to teacher_set_books

LucasCioffi:
Using a standard format for the name of this join table will make it easier to clean up the database.

### DIFF
--- a/app/models/book.rb
+++ b/app/models/book.rb
@@ -8,8 +8,8 @@ class Book < ActiveRecord::Base
   # attr_accessor :matching_api_items
 
 	# has_and_belongs_to_many :authors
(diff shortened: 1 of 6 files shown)


LucasCioffi:
@nnat425 a teacher set is comprised of one or more books.  The `teacher_set_books` table is the join table between those two tables, and it has a column for `book_id` and `teacher_set_id`.

In the book.rb model: 
```
has_many :teacher_set_books, :dependent => :destroy
has_many :teacher_sets, through: :teacher_set_books
```

nnat425:
Interesting, so a book can belong to many different teacher set's. Good to know.

PR #146, opened 2018-09-28
Clean up teacher set records MLN-299

LucasCioffi:
1. Add bnumber to books
2. Index books and teacher sets by bnumber
3. Enforce uniqueness on bnumber in the teacher set and book models and in the indexes
4. Create rake task to do the following:
- delete all duplicate books and teacher sets based on bnumbers (there are no teacher sets with duplicate bnumbers, FYI)
- change the book_id or teacher_set_id of the duplicate associated records (ie holds for teacher sets and teacher sets for books) to match the book or teacher set that wasn't deleted

### DIFF
--- a/app/models/book.rb
+++ b/app/models/book.rb
@@ -11,6 +11,8 @@ class Book < ActiveRecord::Base
 	has_many :teacher_set_books, :dependent => :destroy
 	has_many :teacher_sets, through: :teacher_set_books
 
+  validates_uniqueness_of :bnumber
+
   validate do |book|
     BookValidator.new(book).validate
   end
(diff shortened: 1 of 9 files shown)


LucasCioffi:
@daryanypl just so we close the loop, can you confirm that there's a ticket for re-updating all teacher sets and books in the database in a future sprint?

LucasCioffi:
@daryanypl and @nnat425 if you pull down the latest production database and then run the migrations, the cleanup will occur.  One of the migrations includes the cleanup rake task.

It's interesting to note that when you pull down the production db, you will re-run any migrations that haven't been run on production yet (not a problem, though).  You can see which ones by running `rake db:migrate:status` in the console.

LucasCioffi:
@daryanypl and @nnat425 if you pull down the latest production database and then run the migrations, the cleanup will occur.  One of the migrations includes the cleanup rake task.

It's interesting to note that when you pull down the production db, you will re-run any migrations that haven't been run on production yet (not a problem, though).  You can see which ones by running `rake db:migrate:status` in the console.

daryanypl:
> @daryanypl just so we close the loop, can you confirm that there's a ticket for re-updating all teacher sets and books in the database in a future sprint? You had pointed out that this is necessary today.

Yes, it's https://jira.nypl.org/browse/MLN-301 .

PR #147, opened 2018-09-28
Disable ability to modify teacher sets and books MLN-323

LucasCioffi:
# Before (with buttons)
screen shot 2018-09-28 at 4 20 47 pm
screen shot 2018-09-28 at 4 20 40 pm
screen shot 2018-09-28 at 4 20 32 pm
screen shot 2018-09-28 at 4 20 22 pm

# After (without buttons)
screen shot 2018-09-28 at 4 22 47 pm
screen shot 2018-09-28 at 4 22 42 pm
screen shot 2018-09-28 at 4 22 34 pm
screen shot 2018-09-28 at 4 22 27 pm

### DIFF
--- a/app/admin/books.rb
+++ b/app/admin/books.rb
@@ -1,22 +1,10 @@
 ActiveAdmin.register Book do
+  actions :show, :index
 
   menu :priority => 4
   sidebar :versions, :partial => "admin/version", :only => :show
 
-  controller do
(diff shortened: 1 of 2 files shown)


nnat425:
These images would be more helpful for me if you could also include arrow or marks to point out the noticeable differences.

PR #148, opened 2018-10-02
fix bug where user's can re-cancel orders

nnat425:
Fix minor bug where a user could re-cancel a canceled teacher_set order.

### DIFF
--- a/app/assets/javascripts/controllers/HoldCancelCtrl.js
+++ b/app/assets/javascripts/controllers/HoldCancelCtrl.js
@@ -21,8 +21,8 @@ app.controller('HoldCancelCtrl', [ '$scope', '$routeParams', '$http', '$location
         } else {
            $('#cancel-text').html("This current order of <strong>" + data.teacher_set.title + "</strong> has already beeen cancelled!")
         }
-        $('#cancel-button').prop("disabled",true);
-        $('#cancel-button').css('background-color', 'grey')
+         $('#cancel-button').prop("disabled",true);
+         $('#cancel-button').css('background-color', 'grey')
       }
 
       $scope.hold = data.hold;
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/controllers/HoldCancelCtrl.js
```diff
@@ -16,6 +16,12 @@ app.controller('HoldCancelCtrl', [ '$scope', '$routeParams', '$http', '$location
         return false;
       }
 
+      if (data.hold.status == "cancelled"){
+        $('#cancel-text').html("This current order of <strong>" + data.teacher_set.title + "</strong> has already beeen cancelled!")
```
LucasCioffi: To check for `null`, change this:
`"This current order of <strong>" + data.teacher_set.title + "</strong> has already beeen cancelled!"`
to this:
`"This order is already cancelled."`

There was a typo in "beeen".

daryanypl: Agreed that we need to secure string generation against null.  Another way to do it is to use string interpolation, and check for null or underfined values in there.  There's an example in the holds email generation code.

nnat425: @LucasCioffi @daryanypl Have you seen a case where a teacher set had a title of null?  I checked the db and did not. I think it would be better to add a validation for the teacher_set model preventing any records with the title nil to be inserted. Therefore, avoiding this issue all together.

LucasCioffi: @nnat425 Darya is checking with Courtney about minimum fields.  If the ILS accepts a book created without a title (not sure if possible) and we reject that record from MLN, then the two databases will be out of sync.  Courtney estimates that she might be done with that by the end of today.  Given this, I'd recommend checking for null and we can implement validation based on what Courtney recommends.

nnat425: So it sounds like first, we need to confirm whether the ILS accepts a teacher_set without a title. Second if it does, how are you going to determine the title of the teacher_set? If the ILS does not accept a teacher_set without a title, then the solution I coded will work no problem.

Regardless, I'll check for null here but I believe its redundant code that doesn't need to be here at all.

PR #151, opened 2018-10-02
Require that book bnumbers are unique when present

LucasCioffi:


db/migrate/20181002155016_add_index_to_books.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexToBooks < ActiveRecord::Migration
+  def change
+    remove_index :books, :name => "index_books_bnumber"
+    add_index :books, ["bnumber"], :name => "index_books_bnumber"
```
LucasCioffi: Here we are removing this index and then re-adding (without a uniqueness constraint).

PR #158, opened 2018-10-03
Create and update teacher set records MLN-300, 302, 328

LucasCioffi:
This PR is for accepting JSON related to teacher sets and books.  We use the `bnumber` to look up the teacher set.  If it exists, we update it; if it doesn't exist, we create it.

We also loop through the ISBNs and create associations to those books through the `TeacherSetBook` table.  If the book doesn't exist, we create it.  We also remove teacher_set_books records for books with an ISBN that is not in the teacher_set's list of ISBNs

### DIFF
--- a/app/controllers/api/v1/bibs_controller.rb
+++ b/app/controllers/api/v1/bibs_controller.rb
@@ -1,31 +1,99 @@
 class Api::V1::BibsController < ApplicationController
+  include LogWrapper
+  before_filter :set_request_body
+
+  # Receive teacher sets from a POST request.
+  # All records are inside @request_body.
+  # Find or create a teacher set in the MLN db and its associated books.
   def create_or_update_teacher_sets
-    render status: 200, json: {
-      teacher_sets: [
-        {
-          id: 1,
-          title: 'Example title 1'
-        },
-        {
-          id: 2,
-          title: 'Example title 2'
-        }
-      ]
-    }.to_json
+    error_code_and_message = validate_request
+    if error_code_and_message.any?
+      render_error(error_code_and_message)
+      return
+    end
+
+    saved_teacher_sets = []
+    @request_body.each do |teacher_set_record|
+      teacher_set = TeacherSet.where(bnumber: "b#{teacher_set_record['id']}").first_or_initialize
+      teacher_set.update_attributes(
+        title: teacher_set_record[:title]
+        # TODO: update more teacher_set attributes here once the API contract is complete
+      )
+      teacher_set.update_included_book_list(teacher_set_record)
+      if teacher_set.save
+        saved_teacher_sets << teacher_set
+      else
+        # notify_admin_failed_teacherset(teacher_set)
+      end
+    end
+
+    render status: 200, json: { teacher_sets: saved_teacher_sets_json_array(saved_teacher_sets) }.to_json
   end
 
   def delete_teacher_sets
-    render status: 200, json: {
-      teacher_sets: [
-        {
-          id: 1,
-          title: 'Example title 1'
-        },
-        {
-          id: 2,
-          title: 'Example title 2'
-        }
-      ]
+    error_code_and_message = validate_request
+    if error_code_and_message.any?
+      render_error(error_code_and_message)
+      return
+    end
+
+    saved_teacher_sets = []
+    @request_body.each do |teacher_set_record|
+      teacher_set = TeacherSet.where(bnumber: "b#{teacher_set_record['id']}").first
+      if teacher_set
+        saved_teacher_sets << teacher_set
+        teacher_set.destroy
+      end
+    end
+
+    render status: 200, json: { teacher_sets: saved_teacher_sets_json_array(saved_teacher_sets) }.to_json
+  end
+
+  private
+
+  # build saved_teacher_sets_json_array for the response body
+  def saved_teacher_sets_json_array(saved_teacher_sets)
+    return [] if saved_teacher_sets.empty?
+    saved_teacher_sets_json_array = []
+    saved_teacher_sets.each do |saved_ts|
+      saved_teacher_sets_json_array << { id: saved_ts.id, bnumber: saved_ts.bnumber, title: saved_ts.title }
+    end
+    saved_teacher_sets_json_array
+  end
+
+  # this validates that the request is in the correct format
+  def validate_request
+    if !@request_body.present?
+      return [400, 'request body is missing']
+    end
+    # the tests send data in as params[:_json] but Postman sends it in as an StringIO object
+    @request_body.each do |teacher_set_json|
+      if teacher_set_json['id'].blank?
+        return [400, "ID is missing in teacher_set_json: #{teacher_set_json}"]
+      end
+    end
+    if action_name == 'create_or_update_teacher_sets'
+      # specific methods can go here
+    elsif action_name == 'delete_teacher_sets'
+      # specific methods can go here
+    end
+    return []
+  end
+
+  def render_error(error_code_and_message)
+    LogWrapper.log('ERROR', {
+      'message' => error_code_and_message[1],
+      'method' => "#{controller_name}##{action_name}",
+      'status' => error_code_and_message[0]
+    })
+    render status: error_code_and_message[0], json: {
+      message: error_code_and_message[1]
     }.to_json
+  end
+
+  def set_request_body
+    # binding.pry
+    # request_body = request.body.read
+    @request_body = params[:_json] || JSON.parse(request_body)
   end
 end
--- a/app/models/teacher_set.rb
+++ b/app/models/teacher_set.rb
@@ -55,7 +55,9 @@ class TeacherSet < ActiveRecord::Base
   end
 
   def make_slug
-    self.slug ||= [self.title.parameterize, rand(36**6).to_s(36)].join("-")
+    # check for nil title otherwise parameterize will fail
+    parameterized_title = (self.title || '').parameterize
+    self.slug ||= [parameterized_title, rand(36**6).to_s(36)].join("-")
   end
 
   # Poor man's subject... TODO: replace with column in DB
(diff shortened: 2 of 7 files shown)


app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,20 +1,32 @@
...
+        # TODO: update more teacher_set attributes here once the API contract is complete
+      )
+      teacher_set.update_included_book_list(teacher_set_record)
+      saved_teacher_sets << teacher_set if teacher_set.save
+    end
+
+    # build saved_teacher_sets_json_array for the response body
+    saved_teacher_sets_json_array = []
+    saved_teacher_sets.each do |saved_ts|
+      saved_teacher_sets_json_array << { id: saved_ts.id, bnumber: saved_ts.bnumber, title: saved_ts.title }
```
nnat425: While sending this data back is great, the question becomes do we really need it? If we're going log this information in the MLN logs anyway as well as send the response back to the lambda, I feel like it can be a bit redundant. I'm perfectly ok with a 200 response. Keep in mind, I'm already logging this information when I parse the records from the kinesis stream.

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -28,4 +40,34 @@ def delete_teacher_sets
       ]
     }.to_json
   end
+
+  def request_valid?(method_name)
```
nnat425: Why not separate this validation method to 2 different validation methods? One for delete and the other for update/create.

lib/log_wrapper.rb
```diff
@@ -1,29 +1,26 @@
...
-		'ERROR' => 3,
-		'FATAL' => 4,
-		'UNKNOWN' => 5 
-	   }
+		message['timestamp'] = Time.now.strftime("%Y-%m-%dT%H:%M:%S.%L%z")
+		message['level'] = level
 
-	message['timestamp'] = Time.now.strftime("%Y-%m-%dT%H:%M:%S.%L%z")
-	message['level'] = level
+		Rails.logger.add(severity_levels[level], message.to_json)
```
nnat425: Good catch! Thanks again.

app/models/teacher_set.rb
```diff
@@ -665,4 +667,32 @@ def as_json(opts={})
   end
 =end
 
+  # Recieve JSON related to a teacher_set.
+  # For each ISBN, ensure there is an associated book.
+  # Disassociate books that are no longer in the teacher set.
+  def update_included_book_list(teacher_set_record)
```
nnat425: Method looks great. You can also break the blocks of code inside the method to 2-3 other methods. Something to think about.

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -28,4 +40,34 @@ def delete_teacher_sets
...
+        end
+      end
+    elsif method_name == 'delete_teacher_sets'
+      # TODO: for a future PR
+    end
+    return true unless @error_occurred
+  end
+
+  def render_error(status_code, error_message)
+    LogWrapper.log('ERROR', {
```
nnat425: Make sure you include the module in the class.

app/models/teacher_set.rb
LucasCioffi: Thinking out loud here; I welcome more discussion on it...
I cleaned up the method so that it's more readable now.  Because our team's norm is to add comments to methods, I think there's less of a need to break larger methods into smaller ones.  The primary advantage of doing so is to increase readability (many people create readable method names to serve as code comments), however we've got long comments which are more descriptive than those short method names would be.  

These would be private methods so they wouldn't be tested, so there isn't much of a reliability advantage, and it would be more lines of code in total with a chance that a variable isn't available in the private method's scope if it isn't passed in, so I recommend that we stick with what we've got.

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,31 +1,81 @@
...
-          id: 1,
-          title: 'Example title 1'
-        },
-        {
-          id: 2,
-          title: 'Example title 2'
-        }
-      ]
-    }.to_json
+    return if !request_valid?('create_or_update_teacher_sets')
```
daryanypl: Does it return any sort of a value, to indicate the error?

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,31 +1,81 @@
...
+
+    saved_teacher_sets = []
+    @request_body.each do |teacher_set_record|
+      teacher_set = TeacherSet.where(bnumber: "b#{teacher_set_record['id']}").first_or_initialize
```
daryanypl: why the hash tag after the 'b'?

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,31 +1,81 @@
...
+      teacher_set.update_attributes(
+        title: teacher_set_record[:title]
+        # TODO: update more teacher_set attributes here once the API contract is complete
+      )
+      teacher_set.update_included_book_list(teacher_set_record)
+      saved_teacher_sets << teacher_set if teacher_set.save
```
daryanypl: what happens if .save fails?

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,31 +1,81 @@
...
+      # specific methods can go here
+    end
+    return true unless @error_occurred
+  end
+
+  def render_error(status_code, error_message)
+    LogWrapper.log('ERROR', {
+      'message' => error_message,
+      'method' => "#{controller_name}##{action_name}",
+      'statusCode' => status_code
```
daryanypl: I think I saw Raj using "status", rather than "statusCode".

test/functional/api/v1/bibs_controller_test.rb
```diff
@@ -1,17 +1,1351 @@
...
     @controller = Api::V1::BibsController.new
   end
 
-  test "should create or update teacher sets when given a bib number" do
-    post 'create_or_update_teacher_sets'
+  # test 'respond with 400 if request body is missing' do
+  #   post 'create_or_update_teacher_sets'
+  #   assert response.status == 400
+  # end
+
```
daryanypl: love unit tests!!!!!!!

test/functional/api/v1/bibs_controller_test.rb
```diff
@@ -1,17 +1,1351 @@
...
+  test "respond with 400 if ID isn't present for at least one JSON record" do
+    post 'create_or_update_teacher_sets', { _json: [{ not_an_id: 0 }] }
+    assert response.status == 400
+  end
+
+  test "should create or update teacher sets when given a ID (a.k.a. bnumber)" do
+    post 'create_or_update_teacher_sets', { _json: [{ id: BNUMBER1, another_field: 'x' }, { id: BNUMBER2, another_field: 'x' }] }
```
daryanypl: would this fail without a title now?

test/functional/api/v1/bibs_controller_test.rb
```diff
@@ -1,17 +1,1351 @@
...
+
+  test "should create create associated books" do
+    assert Book.count == 0
+    assert TeacherSet.count == 0
+    post 'create_or_update_teacher_sets', { _json: TWO_TEACHER_SETS_WITH_10_ISBNS_EACH }
+    assert_response :success
+    assert Book.count == 10
+    assert TeacherSet.count == 2
+    assert TeacherSet.first.books.count == 10 # both teacher sets have the same books
+    assert TeacherSetBook.count == 20 # both teacher sets have the same books
```
daryanypl: nice

app/controllers/api/v1/bibs_controller.rb
LucasCioffi: `request_valid?` returns either nil or true.

app/controllers/api/v1/bibs_controller.rb
LucasCioffi: The hashtag is for string interpolation.  `"My name is #{name}"` turns into "My name is Bill".

app/controllers/api/v1/bibs_controller.rb
LucasCioffi: `.save` fails if the record fails validation, and therefore we shouldn't return it in the response body.

test/functional/api/v1/bibs_controller_test.rb
LucasCioffi: Yes, in the next PR we'll include validation.

app/controllers/api/v1/bibs_controller.rb
LucasCioffi: Alrighty @daryanypl I made the change in the latest commit to this and the other two places in the app using `statusCode`.

app/controllers/api/v1/bibs_controller.rb
daryanypl: But the value you return, is it a nil, or underfined?  Basically, is this a third type of possible return value?

app/controllers/api/v1/bibs_controller.rb
```diff
@@ -1,31 +1,83 @@
...
-      ]
-    }.to_json
+    return if !request_valid?('create_or_update_teacher_sets')
+
+    saved_teacher_sets = []
+    @request_body.each do |teacher_set_record|
+      teacher_set = TeacherSet.where(bnumber: "b#{teacher_set_record['id']}").first_or_initialize
+      teacher_set.update_attributes(
+        title: teacher_set_record[:title]
+        # TODO: update more teacher_set attributes here once the API contract is complete
```
daryanypl: Would you like this to be part of the next PR?

LucasCioffi:
@daryanypl I'll merge this since we reviewed and modified it together yesterday.

PR #167, opened 2018-10-10
Add 8 parochial schools and mark as active MLN-334

LucasCioffi:


lib/tasks/ingest.rake
```diff
@@ -463,8 +466,9 @@ namespace :ingest do
         SierraCodeZcodeMatch.create!(sierra_code: sierra_code, zcode: zcode)
       end
       School.all.each do |school|
-        # We will raise the error below to prevent a teacher signing up for a school that isn't represented in Sierra's lookup table
-        raise "A SierraCodeZcodeMatch is missing for the school with this zcode: #{school.code}" unless SierraCodeZcodeMatch.find_by_zcode(school.code) || Rails.env.test?
+        # Then ensure all schools have a SierraCodeZcodeMatch by raising the error below
+        # to prevent a teacher signing up for a school that isn't represented in Sierra's lookup table (would cause an error when they place a hold on a book)
+        raise "A SierraCodeZcodeMatch is missing for school ##{school.id} #{school.name} with this zcode: #{school.code}" unless school.code == 'MLNSTAFF' || SierraCodeZcodeMatch.find_by_zcode(school.code) || Rails.env.test?
```
daryanypl: I think the staff school zcode is actually in the sierra pcode table, I think that's where I got it from.  And we would need it to be valid for the site to work.

PR #175, opened 2018-10-15
Update existing teacher set records MLN-301

LucasCioffi:
Fields we're intentionally not saving:
- language
- series
- publication_date
- statement_of_responsibility
- sub_title
- total_copies
- lexile


app/controllers/api/v01/bibs_controller.rb
```diff
@@ -8,23 +8,50 @@ class Api::V01::BibsController < ApplicationController
   def create_or_update_teacher_sets
     error_code_and_message = validate_request
     if error_code_and_message.any?
+      AdminMailer.failed_bibs_controller_api_request(@request_body, error_code_and_message, action_name)
       render_error(error_code_and_message)
-      return
     end
+    return if error_code_and_message.any?
```
nnat425: Line 11 and 14 have the same conditional. I'd suggest adding the return in the first if statement after line 12.

app/controllers/api/v01/bibs_controller.rb
```diff
@@ -8,23 +8,50 @@ class Api::V01::BibsController < ApplicationController
...
 
     saved_teacher_sets = []
     @request_body.each do |teacher_set_record|
-      teacher_set = TeacherSet.where(bnumber: "b#{teacher_set_record['id']}").first_or_initialize
+      # overwrite @teacher_set_record instance variable so it can be read in the var_field and fixed_field methods
+      @teacher_set_record = teacher_set_record
```
nnat425: Why add a "@" in front of teacher_set_record? Are you trying to access it globally from somewhere else?

app/mailers/admin_mailer.rb
```diff
@@ -0,0 +1,35 @@
+class AdminMailer < ActionMailer::Base
+  LOG_TAG = "AdminMailer"
```
nnat425: What is LOG_TAG ?

app/mailers/admin_mailer.rb
```diff
@@ -0,0 +1,35 @@
...
+  default :from => "no-reply@mylibrarynyc.org"
+
+  # Sends an email to let admins know that a request to create/update a bib(s) failed
+  def failed_bibs_controller_api_request(request_body, error_code_and_message, method_name)
+    begin
+      @action_name = action_name
+      @request_body = request_body
+      @error_code_and_message = error_code_and_message
+      emails = AdminUser.pluck(:email)
+      Rails.logger.debug("#{LOG_TAG}.admin_notification: About to send failed_bibs_controller_api_request email")
```
nnat425: Add the Logger Module for this mailer?

app/controllers/api/v01/bibs_controller.rb
LucasCioffi: @nnat425 yes, I mentioned it in the comment above that line (might be hard to see):
"# overwrite @teacher_set_record instance variable so it can be read in the var_field and fixed_field methods"

app/mailers/admin_mailer.rb
LucasCioffi: I copied the pattern from the other mailers.  It looks like the purpose is to specify in the log which class is creating the log message.

app/mailers/admin_mailer.rb
LucasCioffi: Good call.  I added `include LogWrapper` to all the mailers and replaced `Rails.logger...` with `LogWrapper.log`.

app/controllers/api/v01/bibs_controller.rb
```diff
@@ -51,6 +79,26 @@ def delete_teacher_sets
 
   private
 
+  def var_field(marcTag, merge = true)
+    begin
+      if merge == true
+        @teacher_set_record['varFields'].detect{ |hash| hash['marcTag'] == marcTag }['subfields'].map{ |x| x['content']}.join(', ')
+      else
+        @teacher_set_record['varFields'].detect{ |hash| hash['marcTag'] == marcTag }['subfields'].detect{ |hash| hash['tag'] == 'a' }['content']
```
daryanypl: I don't know if it's always an "a".

app/mailers/admin_mailer.rb
daryanypl: It's my little invention.  The wrpapper class the Platform uses lets you have a "meta" field in the log json.  You can out any fields you want into that meta field.  So I started putting the name of the class or execution path in there.  It's an idea from the Android Studio environment -- when you want to only filter log messages that are coming out of a specific process.  Like, "show me all the debug statements that registering the user is making".

PR #176, opened 2018-10-15
rename routes and add swagger doc

nnat425:


config/routes.rb
```diff
@@ -78,12 +79,8 @@
   devise_for :admin_users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self)
 
-  namespace :api do
-    namespace :v1 do
```
nnat425: Unfortunately rails has a problem with a namespace that includes a dot or period :/ Otherwise, I would of kept this format.

PR #186, opened 2018-10-22
add email notification setting

nnat425:


app/mailers/hold_mailer.rb
```diff
@@ -11,7 +11,7 @@ def admin_notification(hold)
       @hold = hold
       @user = hold.user
       @teacher_set = hold.teacher_set
-      emails = AdminUser.pluck(:email)
+      emails = AdminUser.where(email_notifications:true).pluck(:email)
```
LucasCioffi: Please also add this to the two places where we pluck email addresses in `AdminMailer`.

PR #187, opened 2018-10-22
Improve Lexile and grade level parsing MLN-351

LucasCioffi:


test/functional/api/v01/bibs_controller_test.rb
```diff
@@ -273,6 +273,19 @@
           }
         ]
       },
+      {
```
nnat425: I suggest putting JSON data in a separate file since it's so large. It looks cleaner.

LucasCioffi: Thanks, @nnat425.  I pushed that now.

PR #225, opened 2018-12-06
Added rake task to read output file of MLN users from Sierra, and det…

daryanypl:
…ermine which are not in the MLN database.

### DIFF
--- a/lib/tasks/sync_users.rake
+++ b/lib/tasks/sync_users.rake
@@ -1,0 +1,187 @@
+require 'csv'
+require 'pry'
+
+
+namespace :sync_users do
+
+  # Check how many accounts needed to be fixed in the last month.
+  # Iterate through the users in the Sierra CSV.
+  # For each user, see if they exist in the MLN database.
+  # If they do, and if the MLN user was created within the last month, while the Sierra user has been there for longer than that,
+  # then this is a user account manually fixed by the Library Outreach team.
+  # call like this:  RAILS_ENV=local rake check_user_fixes:check_mismatch['db/migrate/20181128_sierra_mln_user_accounts.csv',2,1,'23333090060508']
+  desc "Check manually made account fixes"
+  task :check_user_fixes, [:file_name, :start, :limit, :barcode]  => :environment do |t, args|
+    # if barcode is set, then only checks that barcode for Sierra-MLN mismatch
+    args.with_defaults(:file_name => nil, :start=> 0, :limit => 0, :barcode => nil)
+
+    csv_start = args.start.to_i
+    csv_limit = args.limit.to_i
+
+    csv_barcode = args.barcode
+
+    # Example:  'db/migrate/20181128_sierra_mln_user_accounts.csv'
+    csv_path = args.file_name
+
+    # \u03EA is a unicode character that will never happen in our Sierra output file.
+    # we need this, because the Sierra output has some junk data that's making the CSV gem trip up.
+    # however, this means that we'll now need to strip the double quotes from around the ingested field values ourselves.
+    CSV.foreach(csv_path, col_sep: "|", quote_char: "\u03EA") do |line|
+      # don't process the header row
+      next if $. == 1
+
+      # if we're only reading a few lines of the csv, then see if we can stop, before reading any further
+      next if (csv_start > 0 && $. < csv_start)
+      break if (csv_limit > 0 && $. >= (csv_start + csv_limit))
+
+      # "P BARCODE"|"EMAIL ADDR"|"PIN"|"EXP DATE"|"PCODE3"|"P TYPE"|"TOT CHKOUT"|"HOME LIBR"|"MBLOCK"|"PCODE4"|"PATRN NAME"|"ADDRESS"|"CREATED(PATRON)"|"UPDATED(PATRON)"
+      # Example:  "23333106701234"|"name@gmail.com"|"Tn/jC7sHXQpfw"|"10-01-2018"|"1"|"153"|"4"|"ea   "|"-"|"895"|"LASTNAME, FIRSTNAME"|"123 MAIN ST, NY 10001"|"05-20-2009 15:23"|"07-02-2018"
+      (barcode, email, pin, expiration, pcode3, ptype, total_checkouts, home_library, manual_block, pcode4, name, address, sierra_created, sierra_updated) = line
+
+      # inore bad data, and move on to the next line
+      next if (barcode.nil?)
+
+      # clean off any leading/trailing spaces and double quotes
+      barcode = barcode.strip.downcase
+      barcode = barcode.chomp('"').reverse.chomp('"').reverse
+      sierra_created = sierra_created.chomp('"').reverse.chomp('"').reverse
+
+      # if a barcode was passed, we're looking for that barcode, and should stop processing once we find it.
+      if (!csv_barcode.nil? && csv_barcode == barcode)
+        csv_start = $.
+        csv_limit = 1
+      end
+
+      # is there a user in the mln database that matches this user from Sierra?
+      mln_user = User.find_by_barcode(barcode)
+      if !mln_user.nil?
+        # Yes, there is.  Now see if the user was created in MLN over the last month, while being old in Sierra.
+
+        # was the user created after a month ago in mln database?
+        if (mln_user.created_at > 1.month.ago)
+          begin
+            date_sierra_created = DateTime.strptime(sierra_created, '%m-%d-%Y %H:%M')
+
+            # was the user created before a month ago in Sierra?
+            if (date_sierra_created < 1.month.ago)
+              puts "manually repaired user=#{barcode}, mln_user.id=#{mln_user.id}, mln_user.created_at=#{mln_user.created_at}, mln_user.updated_at=#{mln_user.updated_at}, sierra_created=#{sierra_created}, sierra_updated=#{sierra_updated}"
+            end
+          rescue => exception
+            puts "broke: #{exception}"
+          end
+        end
+      end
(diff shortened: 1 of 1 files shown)


lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+      # inore bad data, and move on to the next line
+      next if (barcode.nil?)
+
+      # clean off any leading/trailing spaces and double quotes
+      barcode = barcode.strip.downcase
+      barcode = barcode.chomp('"').reverse.chomp('"').reverse
+      sierra_created = sierra_created.chomp('"').reverse.chomp('"').reverse
+
+      # if a barcode was passed, we're looking for that barcode, and should stop processing once we find it.
+      if (!csv_barcode.nil? && csv_barcode == barcode)
```
LucasCioffi: You can use `csv_barcode.present?` instead of `!csv_barcode.nil?` for readability.  `.present?` is true if the value is not nil and not empty quotes.  It's the opposite of `.blank?` which is true if the value is nil or empty quotes.

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+        csv_start = $.
+        csv_limit = 1
+      end
+
+      # is there a user in the mln database that matches this user from Sierra?
+      mln_user = User.find_by_barcode(barcode)
+      if !mln_user.nil?
```
LucasCioffi: `mln_user.present?` can work here.

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+        # Yes, there is.  Now see if the user was created in MLN over the last month, while being old in Sierra.
+
+        # was the user created after a month ago in mln database?
+        if (mln_user.created_at > 1.month.ago)
```
LucasCioffi: Is this rake task going to run automatically at the same time and day of month?

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+        csv_start = $.
+        csv_limit = 1
+      end
+
+      # is there a user in the mln database that matches this user from Sierra?
+      mln_user = User.find_by_barcode(barcode)
+      if !mln_user.nil?
+        # Yes, there is.  Now see if the user was created in MLN over the last month, while being old in Sierra.
+
+        # was the user created after a month ago in mln database?
```
LucasCioffi: I'd recommend rephrasing to this: `was the user created in the MLN database within the past month?`

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+        if (mln_user.created_at > 1.month.ago)
+          begin
+            date_sierra_created = DateTime.strptime(sierra_created, '%m-%d-%Y %H:%M')
+
+            # was the user created before a month ago in Sierra?
+            if (date_sierra_created < 1.month.ago)
```
LucasCioffi: You could save this reference to a month ago as a variable at the top of the rake task (`one_month_ago = 1.month.ago`) so that it's exactly the same everywhere in the rake task.

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+      break if (csv_limit > 0 && $. >= (csv_start + csv_limit))
+
+      # "P BARCODE"|"EMAIL ADDR"|"PIN"|"EXP DATE"|"PCODE3"|"P TYPE"|"TOT CHKOUT"|"HOME LIBR"|"MBLOCK"|"PCODE4"|"PATRN NAME"|"ADDRESS"|"CREATED(PATRON)"|"UPDATED(PATRON)"
+      # Example:  "23333106701234"|"name@gmail.com"|"Tn/jC7sHXQpfw"|"10-01-2018"|"1"|"153"|"4"|"ea   "|"-"|"895"|"LASTNAME, FIRSTNAME"|"123 MAIN ST, NY 10001"|"05-20-2009 15:23"|"07-02-2018"
+      (barcode, email, pin, expiration, pcode3, ptype, total_checkouts, home_library, manual_block, pcode4, name, address, sierra_created, sierra_updated) = line
+
+      # inore bad data, and move on to the next line
+      next if (barcode.nil?)
+
+      # clean off any leading/trailing spaces and double quotes
```
LucasCioffi: These four lines are repeated below.  Can you DRY this up?

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+      barcode = barcode.strip.downcase
```
LucasCioffi: How about `barcode = barcode.gsub('"','').strip.downcase` and `sierra_created = sierra_created.gsub('"','')` instead of the reversing below?

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+        # Yes, there is.  Now see if the user was created in MLN over the last month, while being old in Sierra.
+
+        # was the user created after a month ago in mln database?
+        if (mln_user.created_at > 1.month.ago)
+          begin
+            date_sierra_created = DateTime.strptime(sierra_created, '%m-%d-%Y %H:%M')
+
+            # was the user created before a month ago in Sierra?
+            if (date_sierra_created < 1.month.ago)
+              puts "manually repaired user=#{barcode}, mln_user.id=#{mln_user.id}, mln_user.created_at=#{mln_user.created_at}, mln_user.updated_at=#{mln_user.updated_at}, sierra_created=#{sierra_created}, sierra_updated=#{sierra_updated}"
```
LucasCioffi: Do you need to do anything to the user record in the database, or is printing the info sufficient?

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+          rescue => exception
+            puts "broke: #{exception}"
+          end
+        end
+      end
+    end #foreach sierra row
+  end #check_user_fixes
+
+
+  ## NOTE: This method is not complete.  We're abandoning the task, until we can be sure it's warranted.
```
LucasCioffi: Would you like me to review this too?

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+
+
+namespace :sync_users do
+
+  # Check how many accounts needed to be fixed in the last month.
+  # Iterate through the users in the Sierra CSV.
+  # For each user, see if they exist in the MLN database.
+  # If they do, and if the MLN user was created within the last month, while the Sierra user has been there for longer than that,
+  # then this is a user account manually fixed by the Library Outreach team.
+  # call like this:  RAILS_ENV=local rake check_user_fixes:check_mismatch['db/migrate/20181128_sierra_mln_user_accounts.csv',2,1,'23333090060508']
```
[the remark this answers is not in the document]
daryanypl: moved

lib/tasks/sync_users.rake
```diff
@@ -0,0 +1,187 @@
...
+      # if we're only reading a few lines of the csv, then see if we can stop, before reading any further
+      next if (csv_start > 0 && $. < csv_start)
+      break if (csv_limit > 0 && $. >= (csv_start + csv_limit))
+
+      # "P BARCODE"|"EMAIL ADDR"|"PIN"|"EXP DATE"|"PCODE3"|"P TYPE"|"TOT CHKOUT"|"HOME LIBR"|"MBLOCK"|"PCODE4"|"PATRN NAME"|"ADDRESS"|"CREATED(PATRON)"|"UPDATED(PATRON)"
+      # Example:  "23333106701234"|"name@gmail.com"|"Tn/jC7sHXQpfw"|"10-01-2018"|"1"|"153"|"4"|"ea   "|"-"|"895"|"LASTNAME, FIRSTNAME"|"123 MAIN ST, NY 10001"|"05-20-2009 15:23"|"07-02-2018"
+      (barcode, email, pin, expiration, pcode3, ptype, total_checkouts, home_library, manual_block, pcode4, name, address, sierra_created, sierra_updated) = line
+
+      # inore bad data, and move on to the next line
+      next if (barcode.nil?)
```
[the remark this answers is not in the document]
daryanypl: I know, but I like parentheses, they keep people out of trouble.

lib/tasks/sync_users.rake
daryanypl: I don't know...  The gsub function would strip from anywhere in the string, which is not the intent, but isn't actually a bad thing for a barcode.

lib/tasks/sync_users.rake
daryanypl: nice, thanks

lib/tasks/sync_users.rake
daryanypl: I know the grammar is ugly, but it works clearer with the comment above it...

lib/tasks/sync_users.rake
daryanypl: No, just once.  And I actually won't be running it on production -- I already ran it on a copy of production locally.  I'm just putting the code into the repository for posterity.  

If I'm asked to complete the second portion (the insertion of users automatically into the mln database), then I'll need to worry about running the rake task.

lib/tasks/sync_users.rake
daryanypl: not at this time, I'm making the decision to not complete the functionality requested, because auto-inserting user records is slightly risky, and the results of the task I did run show a too-low number of records to be fixed to justify the risks.

lib/tasks/sync_users.rake
daryanypl: no, thanks.  I might delete this code altogether, for code hygiene, or keep it in case I need to complete.

PR #226, opened 2018-12-10
Add allowed email address masks

daryanypl:


app/models/allowed_user_email_masks.rb
```diff
@@ -0,0 +1,5 @@
+class AllowedUserEmailMasks < ActiveRecord::Base
```
nnat425: What exactly is a mask?

app/views/devise/registrations/_sign_up_async_request.js.erb
```diff
@@ -1,3 +1,5 @@
+console.log("hello world")
```
nnat425: You can remove this.

app/views/devise/registrations/_sign_up_async_request.js.erb
```diff
@@ -108,12 +110,19 @@ function validateEmailFormat(email) {
...
 }
 
-function validateEmailDomain(email){
-	var domain = email.split('@')
-	if(domain[1] != "schools.nyc.gov"){
+
+<% allowed_email_patterns = AllowedUserEmailMasks.where(active:true).pluck(:email_pattern) %>
+function validateEmailDomain(email) {
+	var domain = email.split('@');
+  var allowed_email_patterns = <%= raw allowed_email_patterns %>;
```
nnat425: Fix indentation of line 116 and line 117

app/views/devise/registrations/new.html.erb
```diff
@@ -20,7 +20,9 @@
 </noscript>
 
 <h1>Sign Up</h1>
-<%= javascript_include_tag('sign_up_async_request.js') %>
+<script>
+  <%= render :partial => 'sign_up_async_request.js.erb', :locals => { :my_thing => @resource } %>
```
nnat425: What is my_thing?

daryanypl: oops

app/models/allowed_user_email_masks.rb
daryanypl: It's a regular expression that can be used to isolate/approve a string.  I'm not using regex, just the simple form of "@nypl.org is allowed now".  I wasn't sure the best way to name it -- "domain" seems not accurate...

nnat425: Maybe add a comment re-iterating that? Thoughts?

daryanypl: Fair.

daryanypl: done

app/admin/allowed_user_email_masks.rb
```diff
@@ -0,0 +1,6 @@
+ActiveAdmin.register AllowedUserEmailMasks do
+  menu :priority => 100
```
LucasCioffi: The new Ruby syntax is `priority: 100` instead of `:priority => 100`.
Does it work here and everywhere else you added a `=>`?

app/models/user.rb
```diff
@@ -17,11 +17,13 @@ class User < ActiveRecord::Base
   # Makes getters and setters
   attr_accessor :pin
 
+
   # Validation's for email and pin only occurs when a user record is being
   # created on sign up. Does not occur when updating
   # the record.
   validates :school_id, :first_name, :last_name, :presence => true
-  validates_format_of :email, with: /\@schools.nyc\.gov/, message: ' should end in @schools.nyc.gov', :on => :create
+  validate :validate_email_pattern, :on => :create
```
LucasCioffi: Should we also validate on update too?

app/models/user.rb
```diff
@@ -39,6 +41,28 @@ class User < ActiveRecord::Base
...
+  # it doesn't currently pose a problem.
+  def validate_email_pattern
+    if (!defined?(email) || (email == nil) || !email.index('@'))
+      errors.add(:email, 'is required and should end in @schools.nyc.gov or another participating school address')
+      false
+    end
+
+    allowed_email_patterns = AllowedUserEmailMasks.where(active:true).pluck(:email_pattern)
+
+    index = email.index('@')
```
LucasCioffi: You could do `domain = email.downcase.strip.split('@')[-1]` which will return a string even if `@` isn't present.

Downcase and strip help normalize your data.  You may want to add them to after_create and after_update on the AllowedUserEmailMask model too.

app/views/devise/registrations/_sign_up_async_request.js.erb
```diff
@@ -108,12 +108,19 @@ function validateEmailFormat(email) {
 	}
 }
 
-function validateEmailDomain(email){
-	var domain = email.split('@')
-	if(domain[1] != "schools.nyc.gov"){
+
+<% allowed_email_patterns = AllowedUserEmailMasks.where(active:true).pluck(:email_pattern) %>
+function validateEmailDomain(email) {
+	var domain = email.split('@');
```
LucasCioffi: Do you want to get `.split('@')[-1]` then just `domain` below on line 118.

test/unit/user_test.rb
```diff
@@ -66,11 +66,14 @@ class UserTest < ActiveSupport::TestCase
...
 
+
+  ## NOTE:  We manage allowed email addresses dynamically now, but schools.nyc.gov
+  # is to always be allowed, and it's OK not to test registration success for the others.
   [generate_email_without_valid_domain].each do |new_email|
     test 'should not save user without schools.nyc.gov domain in email' do
       @user.email = "testing@gmail.com"
       @user.save
-      assert_equal([" should end in @schools.nyc.gov"], @user.errors[:email])
+      assert_equal([" should end in @schools.nyc.gov or another participating school address"], @user.errors[:email])
```
LucasCioffi: Should you also create a test with a second domain other than `schools.nyc.gov`?

app/models/user.rb
daryanypl: good idea

app/views/devise/registrations/_sign_up_async_request.js.erb
daryanypl: The domain formula would return an empty string, if thing go wrong, and that would match everything.  So would need tweaking.

app/models/user.rb
daryanypl: made a few safety fixes around these lines

test/unit/user_test.rb
daryanypl: Maybe?  What would you want the test to be testing for?  Maybe a success on inserting user with non DOE address?  (The user insert with DOE address is currently going by same mechanism as the non-DOE would, though...)

README.md
```diff
@@ -110,6 +110,10 @@ In many rails projects when you run the server with `rails s` Rails sets RAILS_E
 
 Testing
 ========================
+First, set up a test database:
+bundle exec rake db:drop RAILS_ENV=test
+bundle exec rake db:create RAILS_ENV=test
```
nnat425: I think its important note that we're using a postgres database. The reason is, if your postgres test database isn't created, I think there might be possible issues when executing line 115. I'm not too certain but I think it would be a good idea to try the following process: Execute line 115 and make sure there isn't an already created db called test. What are the results?

test/unit/user_test.rb
LucasCioffi: You could test creating an AllowedUserEmailMask and see if a user can register with that.

PR #247, opened 2019-05-14
[MLN-405] - added availability count and total count

gonuguntla:
* Updated MLN Set Page With Availability (FE display from MLN DB)

### DIFF
--- a/app/controllers/teacher_sets_controller.rb
+++ b/app/controllers/teacher_sets_controller.rb
@@ -66,7 +66,9 @@ class TeacherSetsController < ApplicationController
     render json: {
       :teacher_set => @set,
       :active_hold => @active_hold,
-      :user => current_user
+      :user => current_user,
+      #Display's only  status 'new' holds count in teacher set page.
+      :holds => @set.holds.collect{|i| i.status  if i.status == 'new'}.compact.length
       # :teacher_set_notes => @set.teacher_set_notes,
       # :books => @set.books
     }, serializer: TeacherSetForUserSerializer, root: "teacher_set"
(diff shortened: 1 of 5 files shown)


app/controllers/teacher_sets_controller.rb
```diff
@@ -66,7 +66,9 @@ def show
     render json: {
       :teacher_set => @set,
       :active_hold => @active_hold,
-      :user => current_user
+      :user => current_user,
+      #Display's only  status 'new' holds count in teacher set page.
```
daryanypl: Maybe rephrase to something like:
\# Count the number of new holds on this teacher set, and pass through so the front end can estimate availability.

app/controllers/teacher_sets_controller.rb
```diff
@@ -66,7 +66,9 @@ def show
...
+      :holds => @set.holds.collect{|i| i.status  if i.status == 'new'}.compact.length
```
daryanypl: consider using a more descriptive variable name, like new_holds_count

public/templates/teacher_sets/detail.html
```diff
@@ -6,8 +6,11 @@
       <div class="ts-banner subject subject-{{ts.subject_key}}">
         <i class="fa"></i>
         <h3>{{ts.title}}</h3>
-      </div> 
+      </div>
+      
       <div class="ts-actions">
+        <!-- Display's Teacher set available copies, totatl copies and status 'new' holds count  -->
```
daryanypl: How about 
Displays the number of available copies, total number of copies in stock, and number of recent not-yet-processed holds.

PR #265, opened 2019-05-23
[MLN-380] - implmented  items controller and api endpoints

gonuguntla:


app/controllers/api/v01/items_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  before_filter :set_request_body
+  before_filter :validate_source_of_request
+
+  def update_availability
+    begin
+      http_status = 200
+      error_code_and_message = validate_request
+      if error_code_and_message.any?
+        AdminMailer.failed_items_controller_api_request(@request_body, error_code_and_message, action_name).deliver
```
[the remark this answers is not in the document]
gonuguntla: AdminMailer 'failed_items_controller_api_request'  method already logging errors.

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
...
+    message = (exception && exception.message ? exception.message[0..200] : 'exception or exception message missing')
+    backtrace = (exception ? exception.backtrace : 'exception missing')
+    LogWrapper.log('ERROR', {
+      'message' => "#{message}...\nBacktrace=#{backtrace}.",
+      'method' => method
+    })
+  end
+
+  def api_response_builder(http_status, http_response)
+    render status: http_status, json: http_response
```
daryanypl: do you need to do 
json: {http_response}.to_json
?

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
...
+      return [400, "Parsing error: #{@parsing_error}"]
+    elsif !@request_body || @request_body.empty?
+      return [400, "Request body is empty."]
+    end
+    return []
+  end
+
+  # Requests to the MLN teacher set-updating api must come from our verified lambdas,
+  # unless are being tested or developed.
+  def validate_source_of_request
```
daryanypl: let's put this method into the private block, like it was in the bib_controller

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
   def unauthorized
     render json: { 'message': 'Unauthorized message from MLN Api::V01::GeneralController' }, status: 401
   end
+
+  # set the @request_body instance variable so it can be used in other methods; 
+  #check for parsing errors.
+  def set_request_body
```
daryanypl: let's put this method in the private block, like in the bibs_controller

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
...
+    begin
+      @request_body = params[:_json] || JSON.parse(request.body.read)
+    rescue => e
+      @parsing_error = e
+    end
+  end
+
+  # this validates that the request is in the correct format
+  def validate_request
```
daryanypl: move to private

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
...
+       'method' => 'validate_source_of_request',
+       'status' => 'start',
+       'dataSent' => "request.headers['X-API-Key']:#{request.headers['X-API-Key']}"
+      })
+
+    redirect_to '/api/unauthorized' unless Rails.env.test? || Rails.env.local? || request.headers['X-API-Key'] == ENV['API_GATEWAY_HEADER_KEY']
+  end
+
+  # log the error and render it back to the lambda
+  def render_error(error_code_and_message)
```
daryanypl: private block

app/controllers/api/v01/general_controller.rb
```diff
@@ -4,4 +4,66 @@ class Api::V01::GeneralController < ApplicationController
...
+    LogWrapper.log('ERROR', {
+      'message' => error_code_and_message[1],
+      'method' => "#{controller_name}##{action_name}",
+      'status' => error_code_and_message[0]
+    })
+    return api_response_builder(error_code_and_message[0], error_code_and_message[1])
+  end
+
+  # Prepare and write an error message to the application log.
+  def log_error(method, exception)
```
daryanypl: same here, let's move to private

app/controllers/api/v01/items_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Api::V01::ItemsController < Api::V01::GeneralController
+  include LogWrapper
+
+  before_filter :set_request_body
+  before_filter :validate_source_of_request
+
+  #Here Updating the available_copies, total copies in Teacher Set table.
```
daryanypl: Would you be able to expand the doc with what data you're expecting, what you will do with that data, and what you will do on unexpected behavior or error?

app/controllers/api/v01/items_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def update_availability
+    begin
+      http_status = 200
+      error_code_and_message = validate_request
+      if error_code_and_message.any?
+        AdminMailer.failed_items_controller_api_request(@request_body, error_code_and_message, action_name).deliver
```
daryanypl: would you be able to add the item to the method call, and you can call it with a nil value, like we do with bib controller errors?

app/controllers/api/v01/items_controller.rb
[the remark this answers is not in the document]
daryanypl: OK, then how about something like
LogWrapper.log('DEBUG', {'message' => 'update_availability.start','method' => 'item_controller.update_availability'})
at the method start?

app/controllers/api/v01/items_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        render_error(error_code_and_message)
+      end
```
daryanypl: Do you need a 
`return if error_code_and_message.any?`
?

app/controllers/api/v01/general_controller.rb
gonuguntla: @daryanypl this is common method to build  api response. 
render status: http_status, json: http_response

app/controllers/api/v01/items_controller.rb
gonuguntla: I didnt see any use case with this  line of code. So i didnt added return if error_code_and_message.any?

app/controllers/api/v01/items_controller.rb
daryanypl: How would you feel about something like this:
# Updates the available_copies, total_copies, and availability fields on a teacher_set.
# Receives a list of items from a POST request, each item represented by a JSON record.
# All records are inside @request_body.
# Parses the item bodies to retrieve the bib id.  For each valid item, sends a request to the Bib Service for all of the items belonging to that bib.
# For each item in the returned list, parses the JSON body, to retrieve the due_date.  If the due_date is not null, counts the item as "unavailable".  
# Adds up the total number of items, and the number of items available, and updates the teacher_set fields accordingly.
# On error finding local data, writes a message to the error log, but returns a success to the calling lambda.  
# On error communicating with the Bib Service, returns a failure to the calling lambda (triggering a re-try).

gonuguntla: fine

app/models/teacher_set.rb
```diff
@@ -789,4 +789,10 @@ def update_notes(teacher_set_notes_string)
       TeacherSetNote.create(teacher_set_id: self.id, content: note_content)
     end
   end
+
+  #Updating available count, total count and availability_string
+  def update_available_and_total_count(total_ct, available_ct)
+    availability_string = (available_ct.to_i > 0) ?  AVAILABILITY_LABELS['available']  : AVAILABILITY_LABELS['unavailable']
```
daryanypl: The current values in the database are "available" and "unavailable".  I think the "Available" and "Checked Out" are used on the front end.  If you'd like to be calling a constant here, you'll need to make another one, or use the AVAILABILITY_LABELS keys.

PR #285, opened 2019-06-13
Merging from development to qa

gonuguntla:


test/functional/api/v01/items_controller_test.rb
```diff
@@ -16,25 +16,22 @@ class Api::ItemsControllerTest < MiniTest::Test
   describe '#test update availability method' do
     it 'test error message for update availability; bnumber not found in MLN db' do
       resp = nil
-      error_message = [404, "bibIds are not found in MLN DB."]
-      @mintest_mock1.expect(:call, [nil])
+      error_message = [404, "BIB id is empty."]
```
daryanypl: I think the previous message was more clear and informative.  Did the functionality change to indicate focus on empty var?

gonuguntla: This is test cases only. 

Its there in code.

      unless t_set_bnumber.present?
        render_error([404, "BIB id is empty."])
        return
      end
      unless nypl_source.present?
        render_error([404, "NYPL source is empty."])
        return
      end
      teacher_set = TeacherSet.find_by_bnumber("b#{t_set_bnumber}")
      unless teacher_set.present?
        render_error([404, "BIB id not found in MLN DB."])

PR #289, opened 2019-06-13
reverted papertrail version checks

gonuguntla:


gonuguntla:
@daryanypl, I am testing all the ways.
Build always failing with one test case with existing code.

PaperTrail::Version  getting 2 ways. So build is failing.

PR #291, opened 2019-06-17
[MLN-383]make available button enable and disable based on available count

gonuguntla:


public/templates/teacher_sets/detail.html
```diff
@@ -13,13 +13,13 @@ <h3>{{ts.title}}</h3>
         <div id='total_available_copies' align="center" style="font-weight:bold"> {{ts.available_copies}} of 
           {{ts.total_copies}} Available | {{ts.new_holds_count}} sets pending
         </div><p style="white-space: pre"></p>
-        <div data-ng-show="ts.available_copies > 0 && !active_hold && is_available ">
+        <div data-ng-show="ts.is_copies_available && is_available ">
```
daryanypl: how is is_available different from is_copies_available?

public/templates/teacher_sets/detail.html
```diff
@@ -13,13 +13,13 @@ <h3>{{ts.title}}</h3>
...
           <a ng-click="doReserve()" class="button secondary">{{ts.availability_string}} - Order now!</a>
           <p>Available Teacher Sets will deliver to your school within 2 weeks. For Teacher Sets that are currently in use by other educators, please allow 60 days or more for delivery. If you need materials right away, contact us at <a href="mailto:help@mylibrarynyc.org">help@mylibrarynyc.org</a>.</p>
         </div>
-        <div data-ng-show="active_hold || !is_available">
+        <div data-ng-show="!ts.is_copies_available && !is_available">
```
daryanypl: The requirement from the email thread:
"1) Button is available if there are copies available _and_ button hasn’t already been (successfully) clicked
2) Button is always unavailable if there are zero copies, and also unavailable if button has already been clicked

I think in terms of logic, this is essentially adding the "Button is always unavailable if there are zero copies" on top of the existing logic"

public/templates/teacher_sets/detail.html
gonuguntla: > how is is_available different from is_copies_available?
@daryanypl (is_copies_available  value calculating from available_copies count and is_available calculating  from  availability string)

public/templates/teacher_sets/detail.html
gonuguntla: @daryanypl I tested this functionality  its working as expected . This code for  'unavailble', If we see the this entire  file 'available' code also  there .

public/templates/teacher_sets/detail.html
daryanypl: yes, but that string is itself calculated in the item update code that you wrote before.  and it's dependent on the # of copies available.  basically, I'm not sure you need all three variables.

public/templates/teacher_sets/detail.html
daryanypl: So you're saying that is_available is updated when the teacher places a hold?

gonuguntla: @darya  Yes, For more secure already that field is available in our database do added that condition also. If  not required will remove that condition.

public/templates/teacher_sets/detail.html
```diff
@@ -13,13 +13,13 @@ <h3>{{ts.title}}</h3>
         <div id='total_available_copies' align="center" style="font-weight:bold"> {{ts.available_copies}} of 
           {{ts.total_copies}} Available | {{ts.new_holds_count}} sets pending
         </div><p style="white-space: pre"></p>
-        <div data-ng-show="ts.available_copies > 0 && !active_hold && is_available ">
+        <div data-ng-show="ts.is_copies_available">
```
gonuguntla: is_copies_available condition is required for to show different text message for  available button and unavailable button.
 eg:

AVAILABLE - ORDER NOW!
Available Teacher Sets will deliver to your school within 2 weeks. For Teacher Sets that are currently in use by other educators, please allow 60 days or more for delivery. If you need materials right away, contact us at help@mylibrarynyc.org.


0 of 2 Available | 7 sets pending
UNAVAILABLE
You have placed a hold on this Teacher Set. If you need multiple copies of this Teacher Set, contact us at help@mylibrarynyc.org.

gonuguntla:
there are four distinct cases:
1) 0 available, user not clicked button — BUTTON DISABLED
2) 0 available, user has clicked button — BUTTON DISABLED
*3) >0 available, user not clicked button — BUTTON ENABLED*
4) >0 available, user already clicked button — BUTTON DISABLED


These cases passed with current functionality, gave quick demo to Stephen. This PR is not required.

Closing this PR.

PR #315, opened 2019-07-16
Merging from dev to QA

gonuguntla:


app/controllers/teacher_sets_controller.rb
```diff
@@ -59,20 +59,32 @@ def show
...
+
     if @set.held_by? current_user
       @active_hold = @set.pending_holds_for_user(current_user).first
     end
 
-    max_copies_requestable = ENV['MAXIMUM_COPIES_REQUESTABLE'] || 5
-    if max_copies_requestable.to_i >= @set.available_copies.to_i
+    #Calculates max ordered teacher set quantity from holds table.
+    ts_holds_count = @set.teacher_set_holds_count
+    is_ordered_max_quantity = true if ts_holds_count.to_i >= max_copies_requestable.to_i
```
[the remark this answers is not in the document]
gonuguntla: @nonword If we order max quantity , but should be disabled. We should not allow to order.

app/controllers/teacher_sets_controller.rb
```diff
@@ -59,20 +59,32 @@ def show
...
+
+    #Teacher set available copies less than configured value, we should show ts available_copies count in teacherset order dropdown.
+    if max_copies_requestable.to_i >= @set.available_copies.to_i && ts_holds_count.to_i <= 0
       max_copies_requestable = @set.available_copies.to_i
+    elsif ts_holds_count.to_i >= 0
+      max_copies_requestable = [max_copies_requestable - ts_holds_count.to_i, @set.available_copies.to_i].min
```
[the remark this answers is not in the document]
gonuguntla: @nonword  above condition will not satisfy some cases.

app/controllers/teacher_sets_controller.rb
```diff
@@ -59,32 +59,32 @@ def show
...
-    end
+    max_copies_requestable = [max_copies_requestable - ts_holds_count.to_i, @set.available_copies.to_i].min
 
     render json: {
       :teacher_set => @set,
       :active_hold => @active_hold,
       :user => current_user,
       :allowed_quantities => (1..max_copies_requestable.to_i).to_a,
-      :is_ordered_max_quantity => is_ordered_max_quantity
+      :is_ordered_max_quantity_ts => is_ordered_max_quantity_ts
```
nonword: One suggestion: Rather than compute `is_ordered_max_quantity_ts` in one way and then compute `max_copies_requestable` in a slightly different way, what if you computed `is_ordered_max_quantity_ts` *after* computing `max_copies_requestable` and base it off of that number? (i.e. `is_ordered_max_quantity_ts = max_copies_requestable > 0`).

Also, now that `is_ordered_max_quantity_ts` is scoped to the `user_id`, maybe call it something like `user_has_ordered_max`?

Alternatively, get rid of `is_ordered_max_quantity_ts` altogether and just set `:allowed_quantities` to `[]` when the user has hit their max. That seems semantically pretty clean to me.

Just suggestions with the aim to reduce the number of places the same idea is computed in slightly different ways

app/models/teacher_set.rb
```diff
@@ -59,8 +59,8 @@ def pending_holds_for_user(user)
     end
   end
 
-  def teacher_set_holds_count
-     holds.collect{|i| i.quantity}.inject(:+)
+  def teacher_set_holds_count(user)
```
nonword: As long as we're changing this method .. now that this is scoped to a user, would a better name be `holds_count_for_user`? (The "teacher_set_" prefix is redundant given that it's a member function on a teacher set.)

app/controllers/teacher_sets_controller.rb
```diff
@@ -57,22 +57,32 @@ def index
...
-    end
+    #ts_holds_count is the number of holds currently held in the database for this teacher set.
+    ts_holds_count = @set.holds_count_for_user(current_user)
+    user_has_ordered_max = (ts_holds_count.to_i >= max_copies_requestable.to_i)
+
+    #Teacher set available copies less than configured value, we should show ts available_copies count in teacherset order dropdown.
+    max_copies_requestable = [max_copies_requestable.to_i - ts_holds_count.to_i, @set.available_copies.to_i].min
+
+    #Button should be disabled after teacher has ordered maximum.
+    allowed_quantities = user_has_ordered_max ? [] : (1..max_copies_requestable.to_i).to_a
```
nonword: Love it

app/assets/javascripts/controllers/TeacherSetDetailCtrl.js
```diff
@@ -18,6 +18,7 @@ app.controller('TeacherSetDetailCtrl', ['$scope', '$routeParams', '$http', '$loc
       $scope.active_hold = data.teacher_set.active_hold;
       $scope.ts.teacher_set_notes = data.teacher_set.teacher_set_notes
       $scope.allowed_quantities = data.teacher_set.allowed_quantities;
+      $scope.user_has_ordered_max =  data.teacher_set.allowed_quantities.length <= 0
```
nonword: 👍

PR #328, opened 2019-07-17
Merging from dev to QA

gonuguntla:


app/assets/javascripts/controllers/HoldNewCtrl.js
```diff
@@ -39,7 +39,6 @@ app.controller('HoldNewCtrl', [ '$scope', '$routeParams', '$http', '$location',
       // Otherwise, just submit the form
       } else {
         window.location = '/teacher_sets/' + $scope.teacher_set_id
-        $scope.submitForm();
```
gonuguntla: Here removed submitForam. We do not want to create teacherset hold automatically after login.

PR #332, opened 2019-07-22
Merging from DEV to QA

gonuguntla:
* Displays teacher set holds.


app/controllers/teacher_sets_controller.rb
```diff
@@ -88,6 +88,16 @@ def show
     }, serializer: TeacherSetForUserSerializer, root: "teacher_set"
   end
 
+  #Gets current user teacherset holds from database.
+  def teacher_set_holds
+    @set = TeacherSet.find(params[:id])
+    @holds = @set.holds_for_user(current_user)
```
nodanaonlyzuul: I would drop the `respond_to` block here.
This endpoint is never asked for JSON and the `respond_to` block suggests it is asked for JSON.
Also `resp` is undefined.

PR #365, opened 2019-08-01
Merging from  QA to Dev

gonuguntla:


app/controllers/registrations_controller.rb
```diff
@@ -52,4 +52,25 @@ def after_update_path_for(resource)
...
+  # display  'PIN(here PIN is capital letters) does not meet our requirements. Please try again.'
+
+  def error_msg_hash
+    error_msg_hash = {}
+    if resource.errors.messages[:alt_email].present?
+      error_msg_hash[:alt_email] = ['Alternate email '.concat(resource.errors.messages[:alt_email].join)]
+    end
+
+    if resource.errors.messages[:pin].present?
+      error_msg_hash[:pin] = ['PIN '.concat(resource.errors.messages[:pin].join)]
```
nonword: What's an example value for `resource.errors.messages[:pin]` ?

app/controllers/registrations_controller.rb
```diff
@@ -52,4 +52,25 @@ def after_update_path_for(resource)
...
+    end
+
+    if resource.errors.messages[:email].present?
+      error_msg_hash[:email] = [resource.errors.messages[:email].join]
```
nonword: Presumably the `:email` error message is acceptable as-is? It seems that when there's an error with `:alt_email`, the `errors.messages[:alt_email]` array has the partial sentence "has already been taken". If there's an issue with the primary email, will `errors.messages[:email]` contain a complete sentence like "Email is already in use"?

gonuguntla: @nodanaonlyzuul I forgot to add condition. I  just added for Email also.

PR #376, opened 2019-08-06
Merging from Dev to QA

gonuguntla:


lib/log_wrapper.rb
```diff
@@ -1,26 +1,23 @@
...
+      'UNKNOWN' => 5
+    }
 
-	# Wrapper module server's as a wrapper to
-	# add output to the log file.
-	# The method log take's in 2 parameters.
-	# A string passed in as level and a hash
-	# passed in as message. Please look at
-	# examples from the user.rb model.
+    message['timestamp'] = Time.now.strftime("%Y-%m-%dT%H:%M:%S.%L%z")
```
daryanypl: The reason for the format we're using is so that our logs would be easier to search through on AWS CloudWatch.  See more here: https://github.com/NYPL/engineering-general/blob/master/standards/logging.md

gonuguntla: @daryanypl here  Rubocop identified alignment issue, just i did alignments only,  .Did not change any code.

PR #383, opened 2019-08-09
After login redirect mln457

daryanypl:
Following the logic described here: https://github.com/plataformatec/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update


app/controllers/application_controller.rb
```diff
@@ -1,55 +1,99 @@
 class ApplicationController < ActionController::Base
+
```
gonuguntla: Looks good to me.
Build failed.
170 files inspected, 17 offenses detected
The command "bundle exec rubocop" failed and exited with 1 during .

PR #402, opened 2019-08-21
added quantity of teacher sets ordered to the dashboard page

daryanypl:


app/admin/dashboard.rb
```diff
@@ -106,40 +106,49 @@
...
               end
             end
+
             column 'Call Number', sortable: :call_number do |h|
               if !h.teacher_set.nil?
                 link_to h.teacher_set.call_number, admin_hold_path(h)
               end
             end
+
+            column 'Quantity', sortable: :call_number do |h|
```
gonuguntla: @daryanypl Here * column 'Quantity', sortable: :call_number do |h| *  call number or quantity

app/admin/dashboard.rb
```diff
@@ -181,6 +193,9 @@
             column "#" do i += 1 end
             column 'Set', sortable: :p_set do |h| link_to h.teacher_set.title, admin_hold_path(h) end
             column 'Call Number', sortable: :p_call_number do |h| link_to h.teacher_set.call_number, admin_hold_path(h) end
+            column 'Quantity', sortable: :call_number do |h|
```
gonuguntla: same here also

app/admin/dashboard.rb
daryanypl: hold on, I'm posting an update

PR #414, opened 2019-09-03
Rename queens library mln 472

daryanypl:
renaming the Queens library to "Queens Public Library" everywhere in the sets app.  The task remains to switch the QPL logo image in the page footer -- we're waiting on the image from Queens Marketing department.


app/assets/stylesheets/home.css.scss
```diff
@@ -306,5 +306,13 @@
 footer {
   a img {
     max-height: 80px;
+    max-width: 180px;
   }
 }
+
+#queens_library_logo {
```
gonuguntla: @daryanypl 'queens_library_logo' style commented?

daryanypl: The hash is so we can reference the style by the css element id.
The contents are commented out, because I don't yet have the QPL logo to style.  
I'll take this out, until we get the logo.

PR #434, opened 2019-09-16
MLN-486: made language filter work on both primary_langiage and language colum…

daryanypl:
…ns, and added other languages field to teacher set detail page

### DIFF
--- a/app/models/teacher_set.rb
+++ b/app/models/teacher_set.rb
@@ -160,17 +160,15 @@ class TeacherSet < ActiveRecord::Base
       sets = sets.where("set_type = ?", params[:type])
     end
 
-    if !params[:language].nil? && params[:language].size > 0
+    if params[:language].present?
       sets = sets.where("language IN (?) OR primary_language IN (?)", params[:language], params[:language])
     end
     if !params[:availability].nil? && params[:availability].size > 0
-      sets = sets.where("#{:availability} IN (?)", params[:availability])
+      sets = sets.where("availability IN (?)", params[:availability])
     end
 
     # Sort most available first with id as tie breaker to ensure consistent sorts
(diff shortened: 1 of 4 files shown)


app/models/teacher_set.rb
```diff
@@ -160,10 +160,11 @@ def self.for_query(params)
       sets = sets.where("set_type = ?", params[:type])
     end
 
-    [:language, :availability].each do |prop|
-      if !params[prop].nil? && params[prop].size > 0
-        sets = sets.where("#{prop} IN (?)", params[prop])
-      end
+    if !params[:language].nil? && params[:language].size > 0
```
gonuguntla: instead of !params[:language].nil?  we can use params[:language].present?

app/models/teacher_set.rb
```diff
@@ -160,10 +160,11 @@ def self.for_query(params)
...
+      sets = sets.where("language IN (?) OR primary_language IN (?)", params[:language], params[:language])
+    end
+    if !params[:availability].nil? && params[:availability].size > 0
+      sets = sets.where("#{:availability} IN (?)", params[:availability])
```
gonuguntla: #{} - this is not required for availability, this is column

app/models/teacher_set.rb
```diff
@@ -219,6 +220,7 @@ def self.facets_for_query(qry)
         facets << facets_group
       end
 
+      #binding.pry
```
gonuguntla: Please remove binding.pry

daryanypl: done

app/models/teacher_set.rb
daryanypl: done

app/models/teacher_set.rb
daryanypl: done

PR #438, opened 2019-09-17
[MLN-482] - added pre-k to grades filters

gonuguntla:


public/templates/teacher_sets/list.html
```diff
@@ -17,7 +17,7 @@
         <div class="facetGroup" ng-show="!hideFacet">
           <a href="#/" class="button xsmall clear-filters" ng-show="teacher_sets.facetString">Clear Filters <i class='fa fa-times-circle'></i></a>
           <div class="ui-slider">
-            <h3><a>&nbsp; &nbsp;Grades {{grade_values[0]||'K'}}&ndash;{{grade_values[1]}}</a></h3>
+            <h3><a>&nbsp; &nbsp;Grades {{grade_values[0] == -2? 'Pre K' :  grade_values[0] == -1 ? 'K' :  grade_values[0]}}&ndash;{{grade_values[1] == -2? 'Pre K' :  grade_values[1] == -1 ? 'K' :  grade_values[1]}}</a></h3>
```
gonuguntla: This condition need to write here only , If i write this code in JS file . It will not work.
This file have some link with ng-slider-Grades.js file.

PR #462, opened 2019-09-26
[MLN-494]-Fixed set-type filters

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -159,9 +159,10 @@ def self.for_query(params)
     unless params[:subject].nil?
       sets = sets.where("primary_subject = ?", params[:subject])
     end
-    # Internal name for "Type" is set_type
-    unless params[:type].nil?
-      sets = sets.where("set_type = ?", params[:type])
+    
+    # Internal name for "set type" is set_type
+    unless params['set type'].nil?
```
daryanypl: there's no underscore?  just the space?

gonuguntla: @daryanypl  yes.

PR #487, opened 2019-10-16
Merging from Qa to Master

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -770,9 +772,12 @@ def clean_subject_string(old_subject_string)
 
     # if the subject ends in a period (something metadata rules can require), strip the period
     new_subject_string = new_subject_string.gsub(/\.$/, '').titleize
-    return new_subject_string
-  end
 
+    #New subject string is empty returns nil
+    return unless new_subject_string.present?
```
daryanypl: Let's fix this redundant code before it goes up to prod.

PR #488, opened 2019-10-16
checking nil condition

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -773,10 +773,7 @@ def clean_subject_string(old_subject_string)
     # if the subject ends in a period (something metadata rules can require), strip the period
     new_subject_string = new_subject_string.gsub(/\.$/, '').titleize
 
-    #New subject string is empty returns nil
-    return unless new_subject_string.present?
-
-    return new_subject_string 
+    return new_subject_string if new_subject_string.present?
```
daryanypl: let's try 
return new_subject_string

gonuguntla: @darya new_subject_string is not present do not want to return empty string. user story is remove empty string.

PR #494, opened 2019-10-31
Increase schools code column length

gonuguntla:


db/schema.rb
```diff
@@ -165,7 +165,7 @@
     t.index ["teacher_set_id"], name: "index_book_set_notes_on_book_set_id", using: :btree
   end
 
-  create_table "teacher_sets", force: :cascade do |t|
+  create_table "teacher_sets", id: :bigserial, force: :cascade do |t|
```
gonuguntla: when I did migration id: :bigserial generated automatically

PR #519, opened 2019-11-26
Upgrading rails version fix deprecations

daryanypl:


app/controllers/api/v01/bibs_controller.rb
```diff
@@ -3,8 +3,8 @@
 class Api::V01::BibsController < Api::V01::GeneralController
   include LogWrapper
 
-  before_filter :set_request_body
-  before_filter :validate_source_of_request
+  before_action :set_request_body
```
gonuguntla: @daryanypl  holds controller and application controller still using before filter. Please update those controllers.
class HoldsController < ApplicationController
  include LogWrapper


  before_filter :redirect_to_angular, only: [:show, :new, :cancel]
  before_filter :check_ownership, only: [:show, :update]
  before_filter :require_login, only: [:index, :new, :create, :check_ownership]

daryanypl: done

PR #523, opened 2019-12-03
Fixed rubocop issues

gonuguntla:


.rubocop.yml
```diff
@@ -11,15 +11,9 @@ Layout/AlignHash:
 Layout/EmptyLines:
   Enabled: false
 
-Layout/EmptyLinesAroundClassBody:
```
daryanypl: I like having a few lines after the end of a class body, so that I can scroll better in my IDE.

.rubocop.yml
```diff
@@ -11,15 +11,9 @@ Layout/AlignHash:
...
-  Enabled: false
-
 Layout/EmptyLinesAroundMethodBody:
   Enabled: false
 
-Layout/EmptyLineBetweenDefs:
```
daryanypl: Rubocop insists on too few lines between method declarations.  I like to have 1 blank line separating grouped control blocks, no blank lines between a method's doc text and the method signature, and 2-3 lines between method declarations.  If everything is spaced evenly, then it becomes very difficult to pick method declarations out as I'm quickly scanning through text.

.rubocop.yml
```diff
@@ -29,21 +23,9 @@ Layout/IndentHash:
...
-Metrics/LineLength:
-  Enabled: false
-
-Metrics/MethodLength:
-  Enabled: false
-
 Style/AndOr:
   Enabled: false
 
-Style/BlockDelimiters:
```
daryanypl: I don't have strong feelings about this one.  I disagree with it -- I like block delimiters to be consistent -- you either use all curly braces or all begin-end delimiters.  https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/BlockDelimiters makes it weird for me.  But not weird enough that I couldn't handle it if others feel strongly.

.rubocop.yml
```diff
@@ -29,21 +23,9 @@ Layout/IndentHash:
...
   Enabled: false
 
-Metrics/ClassLength:
-  Enabled: false
-
-Metrics/LineLength:
-  Enabled: false
-
-Metrics/MethodLength:
-  Enabled: false
```
daryanypl: OK, if we're going to enable https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/MethodLength, then let's specify what method length we're shooting for.  How about 
```
Metrics/MethodLength:
    Max: 40
```
and we can scale down from there if we'd like.

.rubocop.yml
```diff
@@ -29,21 +23,9 @@ Layout/IndentHash:
 Layout/SpaceInsideHashLiteralBraces:
   Enabled: false
 
-Metrics/ClassLength:
-  Enabled: false
-
-Metrics/LineLength:
```
daryanypl: How about specifying what line length we're ok with?  Something like:
```Metrics/LineLength:
  Max: 160```
and then we can see what it catches.

File: .rubocop.yml [hunk 6]
```diff
@@ -29,21 +23,9 @@ Layout/IndentHash:
 Layout/SpaceInsideHashLiteralBraces:
   Enabled: false
 
-Metrics/ClassLength:
```
daryanypl: The teacher_set model is 900 lines long.

.rubocop.yml
thisisstephenbetts: I don't get this objection — would be intrigued to understand better. There are other options we can choose — one empty line around the class, etc. Personally I don't like more than 1 as it looks like someone forgot to finish writing the code.

.rubocop.yml
thisisstephenbetts: Again, I prefer consistency (at my last job, my personal preference was 2 lines, but it seemed less important than not surprising my colleagues). If you feel very strongly about it, I don't see a problem with configuring a minimum of 1 and maximum of 2 lines.

.rubocop.yml
thisisstephenbetts: We can add a one-off exception for that model, and refactor, if appropriate, at some later date.

.rubocop.yml
thisisstephenbetts: I use 120, but let's go with 150 — that seems to be a standard.

.rubocop.yml
thisisstephenbetts: Sounds good for now.

.rubocop.yml
thisisstephenbetts: I don't adhere to this one unless I'm checking my code, but it seems like a sensible default.

app/controllers/application_controller.rb
```diff
@@ -24,11 +24,12 @@ def append_info_to_payload(payload)
...
   # Decides where to take the user who has just successfully logged in.
   def after_sign_in_path_for(resource)
-    LogWrapper.log('DEBUG', {'message' => 'after_sign_in_path_for.start','method' => 'app/controllers/application_controller.rb.after_sign_in_path_for'})
+    LogWrapper.log('DEBUG', {'message' => 'after_sign_in_path_for.start',
+                             'method' => 'app/controllers/application_controller.rb.after_sign_in_path_for'})
     # be careful -- after first access stored_location_for clears to a nil, so read it once
     # and store it in a local var before printing out or any other access
     redirect_url = stored_location_for(:user)
-    if !redirect_url.present?
+    unless redirect_url.present?
```
thisisstephenbetts: I'm actually not a fan of this one — I tend to only use inline `unless`. I'm happy to go with it though.

.rubocop.yml
daryanypl: A minimum of 1 would force us to leave space between the method doc and the method signature, which is hard to scan and looks like the method docstring is just some random comment floating about.

A maximum of 2 would not allow me to leave 3 lines, which I prefer and do consistently, especially in codebases with long methods.

app/admin/user.rb
```diff
@@ -51,7 +51,8 @@
       else
         f.input :barcode
       end
-      f.input :pin, label: raw("4-digit PIN<br> It cannot contain a digit that is repeated 3 or more times (0001, 5555) and cannot be a pair of repeated digits (1212, 6363)")
+      f.input :pin, label: raw("4-digit PIN<br> It cannot contain a digit that is repeated 3 or more times (0001, 5555) and \
+                                canccccnot be a pair of repeated digits (1212, 6363)")
```
daryanypl: spelling

app/controllers/api/v01/bibs_controller.rb
```diff
@@ -64,36 +65,47 @@ def create_or_update_teacher_sets
         )
       rescue => exception
         log_error('create_or_update_teacher_sets', exception)
-        AdminMailer.failed_bibs_controller_api_request(@request_body, "One attribute may be too long.  Error: #{exception.message[0..200]}...", action_name, teacher_set).deliver
+        AdminMailer.failed_bibs_controller_api_request(@request_body, 
+                                                       "One attribute may be too long.  Error: #{exception.message[0..200]}...",
```
daryanypl: Let's move these to four spaces right of the start of line -- the extreme indentation is a bit hard to read.

.rubocop.yml
thisisstephenbetts: Think that applies to Layout/TrailingEmptyLines — totally fine with disabling this rule.

.rubocop.yml
thisisstephenbetts: We will have a standard of 2 blank lines for this codebase.

app/controllers/api/v01/bibs_controller.rb
thisisstephenbetts: Let's use 
`Layout/ArgumentAlignment:
    EnforcedStyle: with_fixed_indentation
`

.rubocop.yml
thisisstephenbetts: standardize on 600

thisisstephenbetts:
Hi @gonuguntla,

could you add in the 

```
Layout/ArgumentAlignment:
    EnforcedStyle: with_fixed_indentation
```

statements to rubocop.yml please? I think that will require a few more changes — sorry!

Stephen.

PR #533, opened 2019-12-11
added error logging in addition to email send if being asked to creat…

daryanypl:
…e or update an incomplete bib


gonuguntla:
@daryanypl can you resolve conflicts

PR #541, opened 2019-12-16
[MLN-551]- maintenance-banner & [MLN-550]- Linked image on Teacher Set instance page broken

gonuguntla:


app/views/layouts/_banner.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="alert main-alert fade-in-out-longer" >
```
gonuguntla: Need to change text message

README.md
```diff
@@ -143,9 +143,14 @@ and
...
 ========================
 ```
-MAXIMUM_COPIES_REQUESTABLE :5  - This is a configuration value in AWS ElasticBeanstalk. In future if anyone want to change the value of maximum teacherset orders, we can update in AWS ElasticBeanstalk Configuration.
+MAXIMUM_COPIES_REQUESTABLE: 5  - This is a configuration value in AWS ElasticBeanstalk. In future if anyone want to change the value of maximum teacherset orders, we can update in AWS ElasticBeanstalk Configuration.
 ```
 
+Show Maintenance Banner Configuration
+========================
+```
+SHOW_MAINTENANCE_BANNER: TRUE  - This is a configuration value in AWS ElasticBeanstalk. In future if anyone want to change the value of maintenanct banner, we can update in AWS ElasticBeanstalk Configuration.
```
thisisstephenbetts: This is so that someone can turn on the banner, right? To change the value we have to modify the HTML. So let's say:
"This parameter should be set to the string`true` to turn on the banner, which is set in  app/views/layouts/angular.html.erb and  app/views/layouts/application.html.erb."

Alternatively, we could create a second variable called `MAINTENANCE_BANNER_TEXT` which would actually contain the value to display. That would be useful. Then the sentence would be 
"This parameter should be set to the string`true` to turn on the banner, which is set in  the `MAINTENANCE_BANNER_TEXT` environment variable."

app/controllers/teacher_sets_controller.rb
```diff
@@ -77,8 +77,8 @@ def show
 
     # limits book titles to unique ones to mask the problem we've been having
     # with the teacher_sets detail page puling up duplicate child book records.
-    ts_books = @set.books.select(:cover_uri).distinct
-
+    ts_books = @set.books.select('DISTINCT ON (books.cover_uri) books.cover_uri, books.id, books.title')
```
thisisstephenbetts: Confused why this is in this PR.

app/views/layouts/angular.html.erb
```diff
@@ -16,6 +16,13 @@
 </head>
 <body class="body-<%= controller.controller_name %>-<%= controller.action_name %>">
   <%= render "layouts/navigation" %>
+
+  <% if ENV['SHOW_MAINTENANCE_BANNER'] && ENV['SHOW_MAINTENANCE_BANNER'].to_s.downcase == "true" %>
+    <div id='banner_id'>
```
thisisstephenbetts: Why not have this `div` element in the `app/views/layouts/_banner.html.erb`? Don't really mind, just wondering.

app/views/layouts/application.html.erb
```diff
@@ -17,6 +17,12 @@
 
 <body>
   <%= render "layouts/navigation" %>
+
+  <% if ENV['SHOW_MAINTENANCE_BANNER'] && ENV['SHOW_MAINTENANCE_BANNER'].to_s.downcase == "true" %>
+    <div id='banner_id'>
```
thisisstephenbetts: Same question...

app/controllers/teacher_sets_controller.rb
thisisstephenbetts: Is this related to MLN-550? Maybe add some explanation in these comments as to why. We can chat tomorrow in stand-up too.

gonuguntla:
this PR is not required

PR #542, opened 2019-12-17
Put warning banner on new servers - MLN-551 Maintenance banner text

gonuguntla:


README.md
```diff
@@ -146,6 +146,18 @@ Order Multiple Teacher Sets Configuration
 MAXIMUM_COPIES_REQUESTABLE :5  - This is a configuration value in AWS ElasticBeanstalk. In future if anyone want to change the value of maximum teacherset orders, we can update in AWS ElasticBeanstalk Configuration.
 ```
 
+Show Maintenance Banner Configuration
+========================
+```
+SHOW_MAINTENANCE_BANNER: TRUE  - This is a configuration value in AWS ElasticBeanstalk. In future if anyone want to change the value of maintenanct banner, we can update in AWS ElasticBeanstalk Configuration.
```
thisisstephenbetts: Please change this section to:
```
SHOW_MAINTENANCE_BANNER: TRUE
This parameter can be set in the ElasticBeanstalk AWS console. 
It should be set to the string 'TRUE' to turn on the banner. The banner text is set in the `MAINTENANCE_BANNER_TEXT` parameter, below.

MAINTENANCE_BANNER_TEXT: 'Maintenance banner text' 
This parameter can be set in the ElasticBeanstalk AWS console. 
It should be set to the value which we wish to appear in the banner. It will only appear if the `SHOW_MAINTENANCE_BANNER` parameter above is set to `TRUE`.
```

app/views/layouts/_banner.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if ENV['SHOW_MAINTENANCE_BANNER'] && ENV['SHOW_MAINTENANCE_BANNER'].to_s.downcase == "true" %>
```
thisisstephenbetts: Think we should also make sure that `MAINTENANCE_BANNER_TEXT` isn't empty as well.

PR #555, opened 2019-12-24
Make travis run coverage report

daryanypl:


daryanypl:
Contains old code that should not have been introduced into the source branch.  DO NOT USE!!!

PR #556, opened 2019-12-24
Travis coverage report

daryanypl:


lib/tasks/combined_tests.rake
```diff
@@ -0,0 +1,36 @@
...
+
+    # for now, specify the test file to run.
+    # this will also help automate for travis
+    t.test_files = Rake::FileList[
+      'test/unit/teacher_set_test.rb',
+      'test/unit/ingest_rake_task_test.rb',
+      'test/integration/user_flow_test.rb',
+      'test/functional/exceptions_controller_test.rb',
+      'test/functional/api/v01/bibs_controller_test.rb',
+      'test/unit/user_test.rb', 'test/unit/book_test.rb'
```
gonuguntla: @daryanypl need to add teacher_sets_controller test file or test/functional/api/v01/*.rb.

Need to add all test files

daryanypl: Thesea re the files we're currently using.  Adding the others is needed, and will be another pr.

PR #576, opened 2020-01-21
[DO not Merge] - Rails app need to support multiple domains

gonuguntla:


gonuguntla:
this pr is not required now. All related code merged to QA .

PR #580, opened 2020-01-23
Merging from development to QA (Production logs issue)

gonuguntla:
* Merged code from local to development for testing
* Now logs are generating. Tested in AWS console and  cloud watch logs


config/environment.rb
```diff
@@ -4,7 +4,5 @@
 require File.expand_path('../application', __FILE__)
 
 # Initialize the rails application
-MyLibraryNYC::Application.initialize!
 
-Rails.logger = Logger.new(STDOUT)
```
gonuguntla: We already had ActiveSupport::Logger.new code  in environment(development.rb, qa.rb, production.rb) files. So here its not required this code.

PR #599, opened 2020-02-10
[MLN-592] - Moved info site code  into sets

gonuguntla:


config/routes.rb
```diff
@@ -81,7 +81,14 @@
...
   # just remember to delete public/index.html.
-  root :to => 'home#index'
+
+  # MLN application supports multiple domains.
+  match 'about/participating-schools' => 'info_site#participating_schools', via: [:get], :constraints => { :host => ENV['MLN_URL'] }
+  match 'about/about-mylibrarynyc' => 'info_site#about', via: [:get], :constraints => { :host => ENV['MLN_URL'] }
+  match '/contacts-links' => 'info_site#contacts', via: [:get], :constraints => { :host => ENV['MLN_URL'] }
+
+  root :to => "home#index", :constraints => {:host => ENV['MLN_SETS_URL']}
+  root :to => 'info_site#index', :constraints => {:host => ENV['MLN_URL']}
```
thisisstephenbetts: You could maybe move line 91 to the top of the block starting on line 86. That would keep all the info site routes grouped together. Then, I _think_ you wouldn't need the `ENV['MLN_SETS_URL']` constraint on line 91. And in fact maybe you could get away from defining `ENV['MLN_SETS_URL']` at all.

config/application.yml
```diff
@@ -6,6 +6,8 @@ case ENV['RAILS_ENV']
 
 when 'local'
 
+  ENV['MLN_SETS_URL'] = "dev-sets.mylibrarynyc.local"
+  ENV['MLN_URL'] = "dev-www.mylibrarynyc.local"
```
thisisstephenbetts: Sorry, but could you change `MLN_URL` to `MLN_INFO_SITE_URL` please? It's a bit clearer. Thanks!

config/routes.rb
```diff
@@ -81,7 +81,14 @@
 
   # You can have the root of your site routed with "root"
   # just remember to delete public/index.html.
-  root :to => 'home#index'
+
+  # MLN application supports multiple domains.
+  root :to => "home#index"
```
thisisstephenbetts: Hi Deepika, I think if this line is here then the app will never reach line 87. I think the app needs to check for the constraint _first_, and have this as the general rule afterwards. So I'd move this line, `root :to => "home#index"` to line 92 (i.e., two lines after `match '/contacts-links' => ...` with a blank line inbetween.

Please double check that the routes still work as expected. Actually, could we create some unit tests to check which routes are invoked depending on the HOST header? Thanks!

PR #606, opened 2020-02-18
[MLN-580] Added alphabet anchors to top of Schools list page

gonuguntla:


app/controllers/schools_controller.rb
```diff
@@ -1,10 +1,23 @@
...
 class SchoolsController < ApplicationController
+  layout 'info_site/application'
   
   def index
-    @schools = School.active
-
+    # Group by first letter of the school.
+    @schools = School.active.group_by { |i| i.name[0] }
+    @schools_arr = []
+    @schools.each do |key, value|
```
thisisstephenbetts: Would be great if you used descriptive names instead of `key` & `value` — I don't really know what they are.

app/controllers/schools_controller.rb
```diff
@@ -1,10 +1,23 @@
...
+      school_hash = {}
+      # If school name is start with alphabet letter, school names will display under aplhabet anchor eg: 'A' Academy for Careers (12), Academy(22).
+      # If school name is not start with alphabet letter, school names will display under '#' anchor. eg: '#' 486 newyork school (86),  56test(234).
+      school_hash['schoolname_first_letter'] = key.match?(/[A-Za-z]/) ? key : '#'
+      school_hash['school_names'] = value.collect do |i|
```
thisisstephenbetts: I'd only use `i` for an integer variable. These are properties on schools, right? This logic looks a little odd to me. Couldn't we get the schools in turn and just do `school.code`? It's totally fine if that's not the best way to do it, but I think this approach needs a bit more commenting to make it clear to me. Thanks!

PR #611, opened 2020-02-19
Implemented test cases for info-site routes

gonuguntla:


gonuguntla:
Raised another PR for this code.

PR #619, opened 2020-03-19
Unique barcodes mln588

daryanypl:
resolve non-unique user barcodes by deleting or editing individual user records.
add unique constraint to the barcode column.


db/migrate/20200319195907_clean_duplicate_barcodes.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
+
+class CleanDuplicateBarcodes < ActiveRecord::Migration[5.0]
+  def up
+    @user = User.find_by_id(135728)
```
gonuguntla: @daryanypl  we can simple above query with one line 

eg: User.find(136451).destroy

daryanypl: I'd like to leave the conditional in there -- we won't have the data in the dev or qa dbs.

daryanypl: The loop idea we can consider...  For me, as I'm going through individual user records one by one, fixing them in that order makes sense, and is more readable.

PR #623, opened 2020-04-06
added return statements after render json

daryanypl:
Found a bug when holds new action was not finishing rendering the response body when it should.  There was also a problem with double-rendering on caught exception.

Added incrementally more docs.

### DIFF
--- a/app/controllers/holds_controller.rb
+++ b/app/controllers/holds_controller.rb
@@ -14,7 +14,7 @@ class HoldsController < ApplicationController
     redirect_to root_url
   end
 
-  
+
   # GET /holds/1.json
   def show
     LogWrapper.log('DEBUG', {'message' => 'show.start', 'method' => 'app/controllers/holds_controller.rb.show'})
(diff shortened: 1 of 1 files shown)


app/controllers/holds_controller.rb
```diff
@@ -27,9 +27,10 @@ def show
       teacher_set: @hold.teacher_set,
       teacher_set_notes: @hold.teacher_set.teacher_set_notes
     }, serializer: HoldExtendedSerializer, root: false
+    return
```
gonuguntla, line 30: @daryanypl Can you give me information about the return statement.

thisisstephenbetts:
Hi @daryanypl, I'm not sure what problem this is aiming to fix. I think it would be good to have a test to demonstrate at least what change in behavior this leads to.  My understanding of render is that it only needs to have a return if there is the possibility of having 2 render calls (e.g., https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors, although I'm not mad on the `render action: blah and return` style — would think `render(action: blah) && return` is more rubocop compliant)

PR #640, opened 2020-06-04
Users without alt email have problem requesting a hold

gonuguntla:


app/models/book.rb
```diff
@@ -5,6 +5,8 @@ class Book < ActiveRecord::Base
   # NOTE: The ISBN field in the database refers to a standard number; sometimes this is an ISBN.
   include CatalogItemMethods
   include Oauth
+  include Exceptions
```
gonuguntla, line 8: need to include  Exceptions module in book.rb. 
Otherwise below code will give the error.
raise Exceptions::InvalidResponse, "Invalid status code of: #{response.code}"

PR #653, opened 2020-06-24
[MLN-622] set_type should be taken from Marc field 526 and update set_type value in teacher-set table

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -35,8 +35,9 @@ class TeacherSet < ActiveRecord::Base
 
   AVAILABILITY_LABELS = {'available' => 'Available', 'unavailable' => 'Checked Out'}
   SET_TYPE_LABELS = {'single' => 'Book Club Set', 'multi' => 'Topic Sets'}
-  TOPIC_SET = 'multi'
-  BOOK_CLUB_SET = 'single'
+  TOPIC_SET = 'Topic Set'
+  BOOK_CLUB_SET = 'Book Club Set'
```
daryanypl, line 39: Nice

db/migrate/20200624170002_update_set_type_to_teacher_sets.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class UpdateSetTypeToTeacherSets < ActiveRecord::Migration[5.0]
+  def change
+    TeacherSet.find_each do |ts|
+      set_type = ts.get_set_type(ts.set_type)
```
daryanypl, line 6: let's also provide what to do if the set type field is either null or a value that doesn't conform

gonuguntla: @daryanypl This method is only to convert 'multi' and 'single' values. not for nil values.
If we want to fix set_type values with out nil we need to update BIB ids from sierra or with back ground methods.

PR #658, opened 2020-06-29
[MLN-678] - updated teacher-sets values nil to value

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -628,84 +629,84 @@ def update_books
     # self.save
   end
 
+  # Old and unused.
```
gonuguntla, line 632: old and unused code

app/models/teacher_set.rb
```diff
@@ -901,6 +914,23 @@ def parse_items_available_and_total_count(response)
...
+    set_type = nil
+    if resp.present? && resp.code == 200
+      resp["data"].each do |bib_record|
+        set_type_marctag = bib_record['varFields'].detect { |hash| hash['marcTag'] == '526' }
+        if set_type_marctag.present?
+          set_type = set_type_marctag['subfields'].map { |x| x['content']}.join(', ')
+        end
+      end
+    end
+    set_type
```
daryanypl, line 930: Is there a default set_type?  If you don't get a meaningful response, what would you like to set the type to?  Or would you prefer to keep the column un-filled, and do the default value calculation later, at the elastic search stage?

gonuguntla: @darya  Calling correct endpoint and checking with response code (resp.present? && resp.code == 200)  .Requirement is what ever value is there in 526 we need to retrieve and save.
If 526 field itself not present can you check update_set_type method in teacher_set.rb. Already this scenarios worked in last sprint.

PR #668, opened 2020-07-14
[MLN-682] - added  nil value for invalid user schools

gonuguntla:


app/controllers/api/v01/bibs_controller.rb
```diff
@@ -137,7 +137,7 @@ def delete_teacher_sets
 
     saved_teacher_sets = []
     @request_body.each do |teacher_set_record|
-      # Delete teacher-set record by bib_id
+      # Get teacher-set record by bib_id
```
daryanypl, line 140: how come this change?  it looks like the teacher set is still being destroyed, is there other stuff that happens?

gonuguntla: @daryanypl I updated little more document

PR #680, opened 2020-07-23
Fixed Rubocop Issues

gonuguntla:


daryanypl:
Looks like there are rvm version issues.  Might have to do with Amazon auto-updating its AMIs.  I'll take a look.

PR #687, opened 2020-07-31
Increased test coverage to 46%

gonuguntla:


config/routes.rb
```diff
@@ -15,7 +15,7 @@
     resources :holds
   end
   resources :books
-  resources :schools, :only => [:index]
+  resources :schools, :only => [:index, :create]
```
daryanypl, line 18: Are we adding functionality to create schools?

gonuguntla: @daryanypl  this method created long back.   I am writing test cases.
  def create
    School.create(school_params)
  end

  private

  # Strong parameters: protect object creation and allow mass assignment.
  def school_params
    params.permit(:name, :code, :active, :address_line_1, :address_line_2, :state, :postal_code, :phone_number, :borough)
  end

PR #702, opened 2020-08-12
Sierra user barcode check

daryanypl:
Method on the user model to check if a given barcode is being used in Sierra.


app/models/user.rb
```diff
@@ -122,7 +135,66 @@ def assign_barcode
...
+      LogWrapper.log('ERROR', {
+          'method' => "#{model_name}.check_barcode_uniqueness_with_sierra",
+          'message' => "patron service threw error while looking for user(#{barcode_to_check}) in Sierra",
+          'status' => response.code,
+          'responseData' => response.body
+        })
+      raise Exceptions::InvalidResponse, "Invalid status code of: #{response.code}"
+    end
+
+    return @barcode_found_in_sierra
```
gonuguntla, line 194: Better to return status code also. Like this [@barcode_found_in_sierra, response.code] .Because status code 200 and 409 in both cases return's  this value is true(@barcode_found_in_sierra = true). Caller method do not know about the status.  If status code is 200 means different functionality and status code 409 means different functionality.

app/models/user.rb
```diff
@@ -61,25 +61,38 @@ def validate_email_pattern
   end
 
 
+  def barcode_found_in_sierra
```
gonuguntla, line 64: Without test cases build will not pass

app/models/user.rb
daryanypl: I see what you're saying, but for now, I'd like to keep the business logic encapsulated in this method.  Let's see if we'll need to do more sophisticated response processing outside of this method.

PR #703, opened 2020-08-13
[MLN-688] - Implemented FAQ tab in admin side and faq page in info site

gonuguntla:


app/admin/faq.rb
```diff
@@ -0,0 +1,81 @@
...
+      f.input :answer
+    end
+    f.actions
+  end
+  
+  show question: 
+    proc {
+      faq = Faq.includes(versions: :item).find(params[:id])
+      faq_version = faq.versions[(params[:version].to_i - 1).to_i].reify
+      # if there's a bug with turning the paper trail into an object (with reify) then display the faq instead of a faq version
```
daryanypl, line 49: nice

db/schema.rb
```diff
@@ -61,15 +61,15 @@
     t.text     "sub_title"
     t.text     "format"
     t.text     "details_url"
-    t.string   "publication_date"
-    t.string   "isbn"
-    t.string   "primary_language"
+    t.string   "publication_date",            limit: 255
```
daryanypl, line 64: did you mean to put the string size limits?  this is a problem when the schema.db file is generated on a local system with some versions of postgres.  unfortunately, it's a major inconvenience.  iirc, there's a workaround, otherwise, we just have to make sure to not include the changes into a commit.

gonuguntla: Ok  I will remove it

PR #727, opened 2020-09-02
User validates barcode mln638 [not ready for review]

daryanypl:


app/models/user.rb
```diff
@@ -1,5 +1,8 @@
+# encoding: UTF-8
 # frozen_string_literal: true
 
+require 'pry'
```
gonuguntla, line 4: @daryanypl  can you remove pry here

app/models/user.rb
```diff
@@ -215,6 +118,48 @@ def validate_pin_pattern
   end
 
 
+  # Another piece of code has determined this user is all set to be called
+  # complete and done.  Note: Usually, expect the next step to be the calling of
+  # Patron Service to create the user in Sierra.
+  def save_as_complete
+    #binding.pry
```
gonuguntla, line 125: remove binding.pry

PR #746, opened 2020-09-18
Merging from master to QA

gonuguntla:


thisisstephenbetts (approved):

app/controllers/application_controller.rb
```diff
@@ -99,6 +99,8 @@ def storable_location?
   def redirect_if_old_domain
     if request.host == ENV['MLN_SETS_SITE_HOSTNAME'] && request.fullpath == "/help"
       redirect_to "#{request.protocol}#{ENV['MLN_INFO_SITE_HOSTNAME']}/faq", :status => :moved_permanently
+    elsif request.host == ENV['MLN_SETS_SITE_HOSTNAME'] && request.fullpath == "/"
+      redirect_to "#{request.protocol}#{ENV['MLN_INFO_SITE_HOSTNAME']}/app#/teacher_sets", :status => :moved_permanently
```
thisisstephenbetts, line 103: I've approved already, but in general I think it's best to use the helper functions to generate these URLs. Maybe we should create a maintenance ticket for it.

PR #749, opened 2020-10-01
DONOT-merge - [MLN-473] Implemented newsletter signup page

gonuguntla:


app/controllers/aws_parameter_store_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    begin
+      response = @systems_manager.get_parameter({
+        name: parameter_name,
+        with_decryption: false
+      })
+      LogWrapper.log('INFO', {'message' => "AWS parameter call  #{response}",
+                              'method' => 'get_parameter_store_values'})
+      response_body = response.parameter
+    rescue Aws::Errors::ServiceError => e
+      raise e
```
daryanypl, line 21: Could we put a log line in here?

app/controllers/aws_parameter_store_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+
+  def news_letter_google_sheet_credentials
+    get_parameter_store_values("/#{ENV[RAILS_ENV]}/google-credentials/mylibrarynyc")
+  end
+
+
+  def google_sheet_credentials
+    client_secret_body = news_letter_google_sheet_credentials
+    JSON.parse client_secret_body.value if client_secret_body.present?
```
daryanypl, line 34: does there need to be a try-catch around the parsing, or is there one later up in the code?

app/controllers/aws_parameter_store_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+
+  private
+
+  def create_client
+    client_options = {
+      region: 'us-east-1'
+    }
+    @systems_manager = Aws::SSM::Client.new(client_options)
+    LogWrapper.log('INFO', {'message' => "TEST LOG TEST TEST #{@systems_manager} TEST LOG TEST TEST", 'method' => 'create_client'})
```
daryanypl, line 45: totally cool with testing, just drawing attention for future cleanup

daryanypl (changes requested):
It looks great.  The one overall comment I'd make, is that newsletter text coming from the google spreadsheet might contain illegal characters or other junk/malicious data.  We might want to sanitize it a tiny bit before senging out the newsletter, just in case.

PR #751, opened 2020-10-08
[MLN-709] validate-email-from-google sheets

gonuguntla:


app/assets/javascripts/application.js
```diff
@@ -77,4 +77,28 @@ $( document ).ready(function() {
       $(this).children(".plusminus").text('-');
     }
   });
-});
\ No newline at end of file
+});
+
+
+function hideNewsLetterValidation(email) {
+  if(email  != ""){
```
daryanypl, line 84: Is it possible for email to be null?

app/controllers/application_controller.rb
```diff
@@ -124,4 +124,13 @@ def store_user_location!
     # :user is the scope we are authenticating
     store_location_for(:user, originating_location)
   end
+
+  
+  def clear_flash_messages
```
daryanypl, line 129: Would it be possible to have a quick doc string on this?

app/controllers/news_letter_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  def index
+    validate_news_letter_email_is_valid
+    GoogleSpreadSheet.new.create_news_letter_email_in_google_sheets(params)
+    respond_to do |format|
+      format.html 
+      format.js 
+    end
+    flash[:notice] = "Thanks for subscribing! You will receive an e-mail confirmation shortly."
+  rescue StandardError => e
+    flash[:error] = e.message
```
daryanypl, line 13: Would you mind logging to ERROR level on this?  Or would INFO be more appropriate here?

app/controllers/news_letter_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+
+  def validate_news_letter_email_is_valid
```
daryanypl, line 17: Maybe a quickie one-liner doc string on this.

app/models/google_spread_sheet.rb
```diff
@@ -0,0 +1,42 @@
...
+  # Get google sheet credentials from AwsParameterStore.
+  def google_sheet_client
+    google_credentials = ENV['NEWS_LETTER_GOOGLE_CREDENTIALS']
+    session = GoogleDrive::Session.from_service_account_key(StringIO.new(google_credentials))
+    @spreadsheet = session.spreadsheet_by_title("MLN newsletter signup requests - #{ENV['RAILS_ENV']}")
+    @worksheet = @spreadsheet.worksheets.first
+  rescue StandardError => e
+    LogWrapper.log('ERROR', {'message' => "Error occcured while calling the google sheets. #{e.message}",
+                             'method' => 'google_sheet_client'})
+    raise "We've encountered an error and were unable to confirm your email"
```
daryanypl, line 17: Is the wording of this message something that came down as a business requirement?  I'm thinking maybe we want to let the user know what to do:  do we want them to wait, or refresh, or send a complaint email to someone?

app/models/google_spread_sheet.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  
+  # Create news-letter email into google sheets
```
daryanypl, line 21: I'm glad to see the doc string, but it repeats the method name.  Is there any more explanation or detail you'd like to put in?

app/models/google_spread_sheet.rb
```diff
@@ -0,0 +1,42 @@
...
+  def create_news_letter_email_in_google_sheets(params)
+    google_sheet_client
+    email = params["email"]
+    if email_already_in_google_sheets?(email)
+      LogWrapper.log('DEBUG', {'message' => "Email already saved in google sheets. Email: #{email}",
+                               'method' => 'create_news_letter_email_in_google_sheets'})
+      raise 'Email is already subscribed.'
```
daryanypl, line 28: Is this error message going to be displayed to the user?

daryanypl (changes requested):

app/assets/javascripts/application.js
gonuguntla: @daryanypl  Yes, this is from UI. Without entering if any one will click submit button, it should raise the validation.

app/models/google_spread_sheet.rb
gonuguntla: @daryanypl Still messages  Stephen is going to give. As of now i added some placeholder messages.

app/controllers/news_letter_controller.rb
gonuguntla: Error is good.

app/assets/javascripts/application.js
daryanypl: If it's possible to call this function with a null, I'm thinking we should check for null value as well as an empty string here.

app/controllers/application_controller.rb
daryanypl: When you say "hitting any action", what do you mean?  What are some of the actions you're preparing for?

app/controllers/news_letter_controller.rb
daryanypl: Do you really want to allow test@com?  When you say "news-letter email", do you mean the email address a teacher will sign up for the newsletter with?  Or something else?

daryanypl (changes requested):

app/controllers/application_controller.rb
gonuguntla: @daryanypl controller action method. def test_method end; If any flash messages persisted  just making null.

app/controllers/news_letter_controller.rb
gonuguntla: @daryanypl Yes. I didn't write any custom regular expression for email validation. Stephen asked me use ruby email validation. Check here this is ruby method.  gem 'email_validator',  EmailValidator.valid? params['email']

app/controllers/application_controller.rb
gonuguntla: Screen Shot 2020-10-09 at 1 17 26 PM

gonuguntla: Please check console.log , Email is not null. Email is string "".

app/assets/javascripts/application.js
gonuguntla, line 84: @daryanypl 
Screen Shot 2020-10-09 at 1 17 26 PM

gonuguntla: Email value is empty string only , not null.

app/controllers/application_controller.rb
gonuguntla: Clear flash messages before hitting any controller method.

app/controllers/news_letter_controller.rb
daryanypl: ok

PR #757, opened 2020-10-21
[MLN-720]- Changed default url  in mylibrary nyc app

gonuguntla:


thisisstephenbetts (approved):
Looks good — thanks Deepika!

PR #781, opened 2020-11-02
[MLN-725] Update available copies while cancelling the hold

gonuguntla:


app/controllers/holds_controller.rb
```diff
@@ -122,9 +123,10 @@ def error_message(exception)
 
   def update
     LogWrapper.log('DEBUG', {'message' => 'update.start', 'method' => 'app/controllers/holds_controller.rb.update'})
-
     @hold = Hold.find_by_access_key(params[:id])
-
+    # Update teacher-set available copies while cancelling the hold.
```
thisisstephenbetts, line 127: Is update only used for cancels? If not you don't want to change `available_copies` here, right?

test/functional/holds_controller_test.rb
```diff
@@ -33,6 +33,7 @@ class HoldsControllerTest < ActionController::TestCase
   end
 
   test "should update hold successfully" do
+    sign_in @user
```
thisisstephenbetts, line 36: How did this work before without the user signed in?

thisisstephenbetts (changes requested):
Hi Deepika,
thanks very much for this — basically looks good. Not sure about assuming the update method is only used for cancels. That's worth checking (and commenting if so). Also, I would like to see some more tests, particularly unit tests for {{app/models/teacher_set.rb}}, and checking that the numbers are correct for different methods in {{test/functional/holds_controller_test.rb}}

@thisisstephenbetts Yes Stephen only update method for holds cancellation. I checked more times. I will add it in comments and yes i will add test cases also.

app/controllers/holds_controller.rb
gonuguntla: @thisisstephenbetts  Yes Stephen only update method for holds cancellation. I checked more times. I will add it in comments and yes i will add test cases also.

test/functional/holds_controller_test.rb
gonuguntla: @thisisstephenbetts  
 created pages starting of the page.
 setup do
    @user = holds(:hold1).user
    @hold1 = holds(:hold1)
    @hold2 = holds(:hold2)
    @school_two = schools(:school_two)
  end

app/controllers/holds_controller.rb
gonuguntla: @thisisstephenbetts Yes. I verified more times, when we cancel the hold in sets page, request is coming to update method only.

http://qa-www.mylibrarynyc.org/app#/holds/3861795ae745d91d7b08d3540/cancel

gonuguntla: @thisisstephenbetts  here update method used for cancellation only

test/functional/holds_controller_test.rb
gonuguntla: @thisisstephenbetts  before update method not using this below method
 **@hold.teacher_set.holds_count_for_user(current_user, @hold.id)).**
 I added new functionality to calculate the current-user holds. So added current-user now.

app/controllers/holds_controller.rb
thisisstephenbetts: I'm only confused because underneath it say `if c{:status] == 'cancelled` implying that this method _could_ be used for something else. It might be more future proof if you moved this calcuation inside that block, so it's close (maybe after) the hold cancel (and we don't want to adjust the total if the cancel fails, so that's another reason).

test/functional/holds_controller_test.rb
thisisstephenbetts: Thanks — makes sense.

test/unit/teacher_set_test.rb
```diff
@@ -116,6 +120,49 @@ class TeacherSetTest < ActiveSupport::TestCase
     end
   end
 
+  # Test current-user teacher-set holds quantity count
+  describe "test teacher-set holds count" do
+    # Test-1
```
thisisstephenbetts, line 125: These comments don't really add anything — I think it's better if you take them out. Thanks!

thisisstephenbetts (changes requested):
Thanks for this. Apologies, but I'm asking for a couple more changes — please move the for updates within the 'cancellation' check — more future proof and we don't want to update the count if the cancel fails. There's a note on that in the PR. 

Lastly, and there's not note for this, I'd like to see some tests on the holds_controller, which basically says 
1. TeacherSet1 has N available_copies
2. Some holds on TeacherSet1 cancelled
3. Now TeacherSet1 has N-n available_copies

That kind of thing.

app/controllers/holds_controller.rb
gonuguntla: @thisisstephenbetts  Sorry.yes I moved now.

test/functional/holds_controller_test.rb
```diff
@@ -32,12 +32,34 @@ class HoldsControllerTest < ActionController::TestCase
     assert_response :success
   end
 
-  test "should update hold successfully" do
-    post :update, params: { id: @hold1.access_key, hold_change: {"comment" => "qqq", "status" => "cancelled"}, hold: {status: "MyText"} }
+  test "update teacher-set available_copies and cancel hold successfully" do
+    sign_in @user
+    expected_ts_available_copies  = 4
```
gonuguntla, line 37: added functional test cases.  
Case1: Before cancellation of hold teacher-set available_copies count is 2. Once hold is cancelled,  status also will change to new to cancelled and based on the quantity we are updating the teacher-set available_copies count.

Case2: If hold is not cancelled  status also not changed.

thisisstephenbetts (approved):
Looks good — thanks!

PR #783, opened 2020-11-06
changed name of method that checks barcode with sierra for clarity

daryanypl:
And moved methods dealing with barcodes functionality into their own section in the user model, for better code organization.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -132,7 +132,7 @@ class User < ActiveRecord::Base
         'pcode3' => pcode3,
         'pcode4' => pcode4
       },
-      'barcodes' => [self.barcode.present? ? self.barcode : self.assign_barcode.to_s],
+      'barcodes' => [self.barcode.present? ? self.barcode : self.assign_barcode!.to_s],
       'addresses': [
         {
           'lines': [
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -278,7 +278,7 @@ class User < ActiveRecord::Base
 
   # ################ THE BARCODES SECTION! ################
 
-  def assign_barcode
+  def assign_barcode!
     LogWrapper.log('DEBUG', {
        'message' => "Begin assigning barcode to #{self.email}",
        'method' => "assign_barcode",
@@ -286,7 +286,7 @@ class User < ActiveRecord::Base
        'user' => {email: self.email}
       })
 
-    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).last.barcode
+    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).last.pluck(:barcode)
     self.assign_attributes({ barcode: last_user_barcode + 1})
 
     LogWrapper.log('DEBUG', {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -286,7 +286,7 @@ class User < ActiveRecord::Base
        'user' => {email: self.email}
       })
 
-    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).last.pluck(:barcode)
+    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).pluck(:barcode).last
     self.assign_attributes({ barcode: last_user_barcode + 1})
 
     LogWrapper.log('DEBUG', {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -267,7 +267,7 @@ class User < ActiveRecord::Base
   # could set the user to "pending", and we'll be checking for those here, as well.
   def save_as_pending
     # do we need to fill in a provisional barcode?
-    if !self.barcode.present?
+    unless self.barcode.present?
       self.barcode = self.assign_barcode
     end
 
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -351,4 +251,132 @@ def pcode3
...
+
+    response = HTTParty.get(
+      ENV['PATRON_MICROSERVICE_URL_V01'] + "?barcode=#{barcode_to_check}",
+      headers:
+        { 'Authorization' => "Bearer #{Oauth.get_oauth_token}",
+          'Content-Type' => 'application/json' },
+      timeout: 10
+    )
+
+    case response.code
```
thisisstephenbetts, line 338: Really not a big deal, and there are a dozen valid ways to write this so I'm not asking for a change, but I wouldn't use a case statement here. I'd probably go with something like:

```
statuses_for_found_barcode = [200, 409]
statuses_for_available_barcode = [404]

expected_statuses = statuses_for_found_barcode + statuses_for_available_barcode

@barcode_found_in_sierra = true if statuses_for_found_barcode.include? response.code
  
LogWrapper.log('DEBUG', {
  'method' => "#{model_name}.check_barcode_uniqueness_with_sierra",
  'message' => message_for_response_code(response.code, barcode_to_check)
  'status' => response.code
})

raise Exceptions::InvalidResponse, "Invalid status code of: #{response.code}"  unless expected_statuses.include? response.code
```

A case statement could well be of use in the hypothetical `message_for_response_code` method though.

Something like that would feel a bit more concise and less repetitive on the logging lines, and the named variables would make the intent clearer to me.

As I say, just a suggestion for discussion, rather than a request to change. Please let me know what you think!

app/models/user.rb
```diff
@@ -351,4 +251,132 @@ def pcode3
...
+       'method' => "assign_barcode",
+       'status' => "end",
+       'barcode' => "#{self.barcode}",
+       'user' => {email: self.email}
+      })
+    return self.barcode
+  end
+
+
+  def barcode_found_in_sierra
```
thisisstephenbetts, line 303: Is this still needed? Looks like you don't call it anymore, unless I'm missing something.

thisisstephenbetts (commented):
Hi Darya,
looks good. I've a couple of minor suggestions and thoughts, and one or two requests for clarifications. If you could respond then we should be able to get to approval pretty soon.
Thanks!
Stephen.

app/models/user.rb
```diff
@@ -351,4 +251,132 @@ def pcode3
...
+    end
+
+    self.status = STATUS_LABELS['barcode_pending']
+    self.save
+  end
+
+
+  # ################ THE BARCODES SECTION! ################
+
+  def assign_barcode
```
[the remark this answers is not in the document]
daryanypl: sure

app/models/user.rb
```diff
@@ -351,4 +251,132 @@ def pcode3
...
+    LogWrapper.log('DEBUG', {
+       'message' => "Begin assigning barcode to #{self.email}",
+       'method' => "assign_barcode",
+       'status' => "start",
+       'user' => {email: self.email}
+      })
+
+    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).last.barcode
```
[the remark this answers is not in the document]
daryanypl: Sure, we can do that: https://apidock.com/rails/ActiveRecord/Calculations/pluck

app/models/user.rb
daryanypl: It will be needed in the future, yes.

app/models/user.rb
daryanypl: I agree that logging lines seem repetitive, but we frequently end up needing to perform slightly different logic for each http status.  In this code, we might needing to revise, both for clarity and level of logging, and also for what we consider "found".

app/models/user.rb
thisisstephenbetts: Gotcha. 👍

app/models/user.rb
thisisstephenbetts: Fair enough. As I said, not asking for a change. A couple of observations and one question: I think writing code for what we might need in the future is an anti-pattern of development. Secondly, for me, the most important question is "what is the important code that another engineer should see"; in this case I'd say that it's the different classes of response codes and the `@barcode_found_in_sierra` instance variable. I think in the original those get a little lost in the log messages, and so make a new engineer work harder. It's not a big deal, still not asking for any changes, but something too think about.

test/unit/user_test.rb
```diff
@@ -16,7 +16,7 @@ class UserTest < ActiveSupport::TestCase
   [generate_barcode].each do |barcode|
     test 'sierra user can be found by barcode' do
       mock_check_barcode_request(barcode, '200')
-      response = @user.check_barcode_found_in_sierra(barcode)
+      response = @user.check_barcode_uniqueness_with_sierra(barcode)
       user_would_be_unique_in_sierra = false
```
thisisstephenbetts, line 20: I realise this isn't in your PR, but setting a variable to false and then checking for the negation of it is really hard to read!

thisisstephenbetts (approved):
Added a couple more comments that you might like to look at, but I'm approving this.

PR #786, opened 2020-11-10
Mln636 add pending flag to user

daryanypl:


app/models/user.rb
```diff
@@ -215,6 +115,29 @@ def validate_pin_pattern
   end
 
 
+  # Another piece of code has determined this user is all set to be called
+  # complete and done.  Note: Usually, expect the next step to be the calling of
+  # Patron Service to create the user in Sierra.
+  def save_as_complete!
+    self.status = STATUS_LABELS['complete']
+    self.save
```
thisisstephenbetts, line 123: There seems to be two versions of this method — one with a `!` and one at the end. Given Rails has `save` and `save!` this doesn't seem unreasonable, but I would definitely use the `save!` method in `save_as_complete!` and `save_as_pending!`, and `save` in `save_as_complete` and `save_as_pending`. Or just go with one version, but use the matching Rails `save`. Also I'd move the code together in the file.

app/models/user.rb
```diff
@@ -351,4 +274,150 @@ def pcode3
...
+       'message' => "Begin assigning barcode to #{self.email}",
+       'method' => "assign_barcode",
+       'status' => "start",
+       'user' => {email: self.email}
+      })
+
+    last_user_barcode = User.where('barcode < 27777099999999').order(:barcode).pluck(:barcode).last
+    # no non-nil barcodes found?  this should never happen, but let's make sure we can handle it
+    if last_user_barcode.blank?
+      last_user_barcode = 27777000000001
```
thisisstephenbetts, line 315: This seems odd, to use a hardcoded barcode. Shouldn't we raise an exception? Or save and log an error... or...? If this is placeholder functionality, better to raise a `TODO` message exception for now.

app/models/user.rb
```diff
@@ -351,4 +274,150 @@ def pcode3
...
+  def save_as_complete
+    self.status = STATUS_LABELS['complete']
+    self.save
+  end
+
+
+  # If the user's barcode is not yet finalized, then set its status to
+  # 'pending' and save.  In the future, there may be other conditions that
+  # could set the user to "pending", and we'll be checking for those here, as well.
+  def save_as_pending
```
gonuguntla, line 291: Duplicate methods. Please remove one method

app/models/user.rb
daryanypl: No, we've been allocated a specific barcode range for MLN.  The library lending branches give patrons plastic cards with a barcode, when the patrons register.  The cards are printed ahead of time, and transported to the branches in large-ish batches.  To prevent barcode conflicts, projects that create new barcodes outside of the branches are allocated ranges.  Our range is 27777000000001 to 27777099999999.  SimplyE has another range.  I'll put these numbers into app-level constants, so it's clear they're not mistakes, and can be set and used throughout.

daryanypl: As to using the minimum as default, I'm assuming that this conditon can only arise if MLN doesn't yet have users.  Well, we do have some users outside of our barcode range, so I suppose I should check for that.  Anyway, I would expect this condition to never arise.

app/models/user.rb
daryanypl: Good catch, looks like a merge error.  Corrected.

app/models/user.rb
daryanypl: done

app/models/user.rb
```diff
@@ -286,7 +288,18 @@ def assign_barcode!
...
+    # should prevent the app from working in a visible manner.
+    min_barcode = Integer(ENV['USER_BARCODE_ALLOTTED_RANGE_MINIMUM'])
+    max_barcode = Integer(ENV['USER_BARCODE_ALLOTTED_RANGE_MAXIMUM'])
+
+    last_user_barcode = User.where("barcode < #{max_barcode}").order(:barcode).pluck(:barcode).last
+    # no non-nil barcodes found?  this should never happen, but let's make sure we can handle it
+    if (last_user_barcode.blank? || (last_user_barcode < min_barcode))
+      last_user_barcode = min_barcode
+    end
+
```
thisisstephenbetts, line 302: Hi Darya,
I like the use of ENVs — that's definitely clearer. A few minor points:

- it's a bit more Rubyish to use `.to_i` rather than `Integer()`, but not a big deal
- I think you want `.nil?` not `.blank?` here
- One way of getting a `nil` is if there are no barcodes at all; another though, is if we've run out of barcodes, and I think many developers reading this would assume that was the most likely case (I did, for a sample of 1!) So I would either check for three error cases — no barcodes at all; barcode less than min_barcode; run out of barcodes — and handle each separately; or, at the very least explain better in comments what is happening.

daryanypl: Well, I think blank will cover a nil.  I was using this as a guide:  https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html .

daryanypl: Sure, I'll add a max range check and more docs.

thisisstephenbetts: > Well, I think blank will cover a nil.
Ah, gotcha. Did not realise this was a Rails addition. Thanks!

thisisstephenbetts: Looks good — thanks!

thisisstephenbetts (approved):

PR #796, opened 2020-11-20
[MLN-738] updated set_type value irrespective of isbns

gonuguntla:


thisisstephenbetts (approved):
Looks good to me, Deepika. Thanks!

PR #833, opened 2021-01-07
Update readme with delayed job

daryanypl:


README.md
```diff
@@ -115,6 +115,39 @@ Server
 In many rails projects when you run the server with `rails s` Rails sets RAILS_ENV to "development".  If you do that with this app, you will connect to the development database on AWS (if you have permission to decrypt the value).  Instead, run `RAILS_ENV=local rails s` to start the server and `RAILS_ENV=local rails c` to run the console.
 
 
+Multithreading
```
thisisstephenbetts, line 118: "Multithreading" is the wrong term to use here, I think. This is much more about asynchronous processing, that can happen at some point in the future, rather than "threading" which is about parallel processing (and where the results of the different threads usually are gathered up). 

"Asynchronous processing" might be a less loaded term.

README.md
```diff
@@ -115,6 +115,39 @@ Server
...
+========================
+Our multithreading functionality is done through ActiveJob, which employs DelayedJob on the backend.
+ActiveJob gets better after rails 5.2, so keep in mind that there is some functionality that is not perfect, until we can upgrade rails.
+
+Backstory:  Rails runs on a single thread.  Even when you schedule asynchronous operations, s.a. sending a request with HTTParty, that request is executed on the same thread as the rest of the application.  Rails simulates multithreading with creative use of scheduling.
```
thisisstephenbetts, line 123: I'd rethink this paragraph. Most programming languages would run an HTTP request synchronously, unless explicitly told not to — the next line of code might rely on the result of that so it couldn't, by default, be run in a different thread. I think the point here is that we want this processing to happen asynchronously, in a separate process, not slowing down the main server process. I believe you _could_ actually use threads in a Rails app, but as the process will block until all threads are completed it wouldn't really do what we want.

Also, what does `s.a.` stand for? Not aware of that abbreviation!

README.md
```diff
@@ -115,6 +115,39 @@ Server
...
+
+To put something on its own thread, you have to first create the second thread by starting the second worker on the same EC2 server:
+```
+RAILS_ENV=qa bin/delayed_job start
```
thisisstephenbetts, line 127: This is great — I think it'd be great if you had a simple set of instructions for running locally for development.

README.md
```diff
@@ -115,6 +115,39 @@ Server
...
+
+It's a good idea to use ActiveJob.  Even though using DelayedJob directly can give more powerful functionality, using the ActiveJob intermediary will allow future maintainers to switch implementations without rewriting app code.
+
+How do you call a method asynchronously?  Call it like we do here:
+```
+FindAvailableUserBarcodeJob.perform_later(user: self)
+```
+Perform, and perform_later calls are scheduled to go at some schedule you've set up.  
+At this point, if you have not started your second worker (remember delayed_job start?), your call will go on your regular app stack.  Multithreading will not happen.  If you have a second worker going, your asynchronous call will be put on that second worker's thread.  If you have multiple asynchronous calls going at the same time, they will be put on the second worker thread, but be single-threaded (schedule one after another) within that thread's stack.
```
thisisstephenbetts, line 139: So for example, here, instead of "Multithreading will not happen" I'd say "the method will be processed synchronously".

README.md
```diff
@@ -115,6 +115,39 @@ Server
...
+
+DelayedJob Logging:
+There are two places you'll see information on the scheduled code runs.  One is the ```delayed_jobs``` database table.  Here, you can see the jobs that are scheduled to be run.  Usually, after a job is completed, its row will be removed from the table.  So be aware that you won't see historical jobs in that table.  DelayedJob configuration is in:
+```
+config > initializers > delayed_job_config.rb
+```
+The second place is the log.  Our is here: ``` tail -500f log/delayed_job.log ``` .
+
+Our asynchronous code is currently used for user barcode creation on account create.  During that process, we need to talk to Sierra one or more times, and those calls can take up time and resources.  For more info on barcodes, see https://confluence.nypl.org/display/DIGTL/User+Barcodes .
```
thisisstephenbetts, line 148: I'd put this further up the README, so people understand quickly the "why" before reading about the "how"

thisisstephenbetts (changes requested):
This is a great start. I've made a few suggestions. The main thing is, I think, that "threading" is the wrong term here — I'd stick to "asynchronous processing" or something like that.

Thanks!

README.md
daryanypl: I used the word with purpose.  Asynchronous code is executed in rails in either one of two ways:
- asynchronous process runs in same thread
- asynchronous process runs in second thread, by some third-party magic that rails doesn't provide.

If all we wanted was asynchronous in the same thread, we would have used HttParty, like we do for patron email verification now.  But because we wanted to use multithreading, we worked on this epic.

README.md
daryanypl: HttParty gem runs in same thread, asynchronous execution.  

Sorry, "s.a." stands for "such as".  It's in the same family of use as "ex." ("example") and "etc." ("etcetera").

README.md
daryanypl: I just replace the env with "local".  There are other ways to run the worker, but they're not more useful, and I'd rather keep the instructions easy.

README.md
daryanypl: same as above

README.md
daryanypl: ok

README.md
thisisstephenbetts: In what way is HTTParty asynchronous?

daryanypl: If I understand this correctly, HttParty can be run multithreaded, on the more common RoR implementations, because it's an I/O operation (https://www.rubyguides.com/2015/07/ruby-threads/).  So, if you're using it to talk to Sierra, and the request takes a long time, you're probably going to be fine on one of our AWS Passenger servers.  However, HttParty is designed to be sure to be asynchronous, but not to be explicitly multithreaded.  So, you can have a list of asynchronous operations on the stack, and you check them every x milliseconds to see if it's time to do one.  (There is a good explanation I read a while back, I'm looking for the link.)  So, HttParty is always asynchronous, and usually multithreaded, but it's not guaranteed.  Some tasks are not considered I/O.  For those tasks, RoR doesn't do multithreading by default.  In addition, from what I understand, multithreading still happens within the same process on the CPU.  If you run such a task asynchronously/multithreaded on the same process as the main app, you'll risk slowing down the app.  

If we want to be sure of the multithreaded implementation that doesn't slow down the app and won't crash the app if it crashes, we need to make sure we're spinning our own worker with an ActiveJob.  That worker then spins a process, and there can be one or more threads on that process.

Originally, the functionality we wanted was to periodically crawl through rows of "incomplete" users, attempting to reconcile their barcodes.  I felt that this was a potentially slow and resource-intensive operation (albeit an I/O one), and it'd be safer to follow other Digital groups' example, and use ActiveJob.

thisisstephenbetts: Okay, thanks for that.

Normally when we say a function or a method is asynchronous we mean that the function returns before the action is complete. This is the prevalent mode in node.js for example. But languages like Ruby & Python are synchronous, unless you explicitly say otherwise (using threads, for example).

To illustrate:

```
HTTParty.get 'https://nypl.org/'
puts 'next line'
```
when `next line` is printed the `get` method will have completed.

In JavaScript it tends to be asynchronous:
```
    xmlhttp = new XMLHttpRequest();
    xmlhttp.open("GET", 'https://nypl.org');
    xmlhttp.send(null);

    console.log('next line');
```
so `next line` will be printed _before_ the request has completed.

It's totally true that HTTP requests, like most IO operations, don't block at the OS level, so that they don't consume CPU cycles while they wait — but that is different from it being asynchronous at the application or code level. 

Threads are different — they are created within a single application or executable. So they're suitable for when you have two bits of code that can be executed in parallel, not for something where we want to do somewhat lengthy processing at a later time. It can be confusing, as people sometimes use 'thread' a little bit more generally, but "multithreading" in a technical context should always mean using actual threads at the application level. (As far as I can see HTTParty does not use threads at all.)

Here's a trivial example in Ruby:
```
t1 = Thread.new{func1()}
t2 = Thread.new{func2()}
t1.join
t2.join
```

(There are other ways of achieving asynchronous code too — event-driven execution, as node.js uses, is one.)

It seems to me that the crucial thing here is that we are registering ActiveJobs that a separate process looks for and processes. Those things _could_ use multithreading, but don't need to, and — as far as I can see — in this code are not.

README.md
thisisstephenbetts: Sorry, I wasn't clear. I meant just have a simple set of instructions to run locally like:

```
$ bundle install
$ RAILS_ENV=local bundle exec rails s
then in a separate window run
$ RAILS_ENV=local bundle exec bin/delayed_job start
```
For example. Not sure if I've got those instructions right, but you get the idea.

app/jobs/find_available_user_barcode_job.rb
```diff
@@ -46,7 +47,7 @@ def perform(user: nil, **args)
       # well, we can't be saving this user with a duplicate barcode.
       # ask the user to increment its barcode, and try again.
       if already_there
```
garvita, line 49:
```suggestion
      if already_there
```

app/models/user.rb
```diff
@@ -120,11 +119,21 @@ def validate_pin_pattern
...
-  def send_request_to_patron_creator_service
+  def send_request_to_patron_creator_service(pin_code)
+    LogWrapper.log('DEBUG', {
+       'message' => "user.send_request_to_patron_creator_service: start with self=#{self || 'NA'}",
+       'method' => "#{model_name}.send_request_to_patron_creator_service",
+       'status' => "start"
+    })
+
+    if pin_code.blank?
+      pin_code = self.pin
```
garvita, line 130: What are the circumstances where the passed `pin_code` would be different from `self.pin`?
In the call it's clear these are the same:
`FindAvailableUserBarcodeJob.perform_later(user: self, pin_code: self.pin)`
so it's confusing why you need to pass in the pin separately here.

app/jobs/find_available_user_barcode_job.rb
```diff
@@ -57,20 +58,20 @@ def perform(user: nil, **args)
     # If we succeeded, it's time to tell the User object that its barcode is no longer pending
     if already_there == false
       begin
-        Rails.logger.debug "#{self.class.name}: barcode [#{user.barcode}] is available in Sierra, calling patron creator service."
+        Delayed::Worker.logger.info("#{self.class.name}: barcode [#{user.barcode}] is available in Sierra, calling patron creator service.")
 
         # TODO: on timeouts/exceptions/negative results,
         # don't send request to Patron Service, keep user as pending
-        user.send_request_to_patron_creator_service
+        user.send_request_to_patron_creator_service(pin_code)
```
garvita, line 65: Why would `pin_code` here be different from `user.pin`?

app/jobs/find_available_user_barcode_job.rb
garvita, line 49: `already_there` is not a clear variable name for someone new to the code. Does this refer to the barcode? Is it "there" in Sierra or in MLN?

README.md
daryanypl: We're making a second worker.  That second worker creates a process.  The process then spins a thread.  Within that thread, DelayedJob is single-threaded -- all of the jobs are run in series, not parallel.  However, we do now have two threads -- the main app thread and the thread that DelayedJob's new process made.

This is a lot of back and forth on a small PR.  I'll change the verbiage.

app/jobs/find_available_user_barcode_job.rb
daryanypl: fair point.  the variable is set in the base branch, I'll update it its own PR.

app/jobs/find_available_user_barcode_job.rb
daryanypl: The user object isn't being serialized correctly, when passed to the ActiveJob, and I don't want to introduce the complexity of creating a custom serializer.

app/models/user.rb
daryanypl: Serialization issues.  I'll make it more explicit with a doc string.

app/jobs/find_available_user_barcode_job.rb
garvita: Where is the serialization happening? Is this a known issue? How have ascertained if this being caused due to serialization?

daryanypl: Objects sent to the ActiveJob perform process are serialized as part of the normal functionality.  This serialization is known to not always include all object fields.  One way to fix it is by writing a custom serializer for User, but I believe that would introduce unnecessary complexity.  If there was more than one field, or if there was some extra functionality I needed, I'd opt for the serializer, but in this case, I think passing the pin manually is the better choice.

README.md
```diff
@@ -115,6 +115,62 @@ Server
 In many rails projects when you run the server with `rails s` Rails sets RAILS_ENV to "development".  If you do that with this app, you will connect to the development database on AWS (if you have permission to decrypt the value).  Instead, run `RAILS_ENV=local rails s` to start the server and `RAILS_ENV=local rails c` to run the console.
 
 
+Asynchronous Execution
+========================
+Why: Our asynchronous code is currently used for user barcode creation on account create.  During that process, we need to talk to Sierra one or more times, and those calls can take up time and resources.  We'd like the user to be able to peruse the site while their barcode is being computed.  For more info on barcodes, see https://confluence.nypl.org/display/DIGTL/User+Barcodes .
```
thisisstephenbetts, line 120: > like the user to be able to peruse the site

It's also so that if Sierra is unavailable the response doesn't hang, and we can process the request later.

README.md
```diff
@@ -115,6 +115,62 @@ Server
...
+
+Our asynchronous functionality is done through ActiveJob, which employs DelayedJob on the backend.
+ActiveJob gets better after rails 5.2, so keep in mind that there is some functionality that is not perfect, until we can upgrade rails.
+
+Backstory:  Rails runs on a single thread within a single process.  When you send a request with HTTParty, that request is executed on the same thread as the rest of the application.  Rails will sometimes simulate asynchronicity with creative use of scheduling on IO operations on some of the Rails servers/environments.  But, generally, HttParty will be running on the same thread and process as your app, and will hold up app execution, if it's stuck.
+
+To make code asynchronous, we're making a second worker, which will start its own process, which will spin its own thread:
```
thisisstephenbetts, line 127: I don't think it's accurate that a process will "spin its own thread". If there's only one thread then the thread is the process. 
I think it's clearer to just say "To make code asynchronous, we're making a second worker" and leave it at that.

README.md
```diff
@@ -115,6 +115,62 @@ Server
...
+RAILS_ENV=qa bin/delayed_job start
+```
+
+To run the whole shebang locally, start your app in one terminal tab:
+```
+$ RAILS_ENV=local rails s
+```
+then in a second terminal tab:
+```
+$ RAILS_ENV=local bin/delayed_job start
```
thisisstephenbetts, line 138: Thanks!

thisisstephenbetts (commented):
I've made a couple more suggestions on the docs. It's more minor than previous comments, and I think what you have is okay. I would approve the doc parts of this PR, with those suggestions. Thanks for the updates.

README.md
thisisstephenbetts: @daryanypl, as mentioned in stand-up, if you have a reference for the "worker starts a process which spins up a thread" please link it. It seems like you are drawing a distinction between a worker, a process and a [single] thread. That may well be correct — I'd be hapy to be wrong and learn something — but I don't understand the distinction is.

app/models/user.rb
garvita: Wouldn't it be easier to pass user ID and fetch the object from the DB and use the pin so you can remove this check.  Rather than passing it two different ways (`pin_code` and `self.pin`)  and then having to verify which one contains the value. This would less be confusing for someone new to understand in the future.

README.md
daryanypl: One place is here:
https://github.com/collectiveidea/delayed_job
"# Runs two workers in separate processes.
RAILS_ENV=production script/delayed_job -n 2 start"

This (https://medium.com/@adrianbooth/exploring-rubys-multi-threaded-weirdness-d5dc61883591) is not a direct answer, but I found it helpful in learning about RoR quirks, and in interpreting all of the other information I've been reading on this topic.  It has visual diagrams, as well as verbal explanations, and I'm a visual learner.

README.md
daryanypl: done

app/models/user.rb
daryanypl: Not necessarily.  We might run into some issues with Devise -- I've seen some funky behavior around when, exactly, the User object gets persisted into the database, and what's in the session and the cache.  Plus, passing the User as a serialized "handler" object is how the hook is meant to be run.  In the future, we might want to replace backends, and I'd like to be as standardized as possible.

daryanypl: I'd like to also be able to call this method from anywhere within the app, not just the job.  If I remove this explicit passing and check, my complexity would still be there, just hidden under the covers.  I'd like to make any possible issues be as visible as possible, for future maintenance and upgrades (a newer version of RoR might not have the same issues, and we'll see the "extra" code and clean it).

app/jobs/find_available_user_barcode_job.rb
gonuguntla: @daryanypl  https://blog.saeloun.com/2019/09/11/rails-6-custom-serializers-for-activejob-arguments.html

app/models/user.rb
gonuguntla: @daryanypl  @thisisstephenbetts  Rails 6 introduces custom serializers for objects which are passed as arguments to ActiveJob.

thisisstephenbetts: Hi Darya, I'd left this section of code alone until now to not get in the way of the conversation between you and Garvita. But as I know you were keen to get this resolved I took a look late last week and did a little research. 

I agree that 
```FindAvailableUserBarcodeJob.perform_later(user: self, pin_code: self.pin)```
raises a bit of an alert in my mind when reading the code. It makes the developer worry why it's done that way. 

And then the 

```
     if pin_code.blank? 
         pin_code = self.pin
```

raises another alert, which compounds the first. It's unclear why we are anticipating pin_code being blank; and if user was not serialized properly, it's unclear why `self.pin` would work now. Additionally, it raises a concern that the pin and user _could_ get out of sync (if the PIN was updated before the active job executes, for example). 

You said during your talk, in reply to Leonard, that only the user_id is stored in the active_job table in the DB. So if we know that, why not just pass that explicitly as Garvita suggested? It's the simplest info to store, and very easy to retrieve from the db. I don't see why Devise or the cache or the session would get in the way of that if called explicitly.
Then the method of the job could just start

```
def perform(user_id)
    user = User.find(user_id)
```

which would be clear and mean the `if pin_code.blank?` conditional can be removed.

That would also work if it was run from anywhere in the app, though if you wanted to avoid unnecessarily refetching the user from the database  (I don't think it's a big deal to do that, but if you _do_ want to avoid it) you could do a refactor so that the subsequent code in the perform method is separated out (either as a User method or a separate function). E.g., 

```
def perform(user_id)
    user = User.find(user_id)

    user.obtain_barcode_in_sierra
```

or 

```
def perform(user_id)
    user = User.find(user_id)

    obtain_barcode_for_user(user)
```

I don't think that is really necessary, though.

I think this, as Garvita had suggested, would be an improvement. It simplifies the code (has fewer execution paths and permutations of data) and makes it clearer.

README.md
thisisstephenbetts: Hi Darya, as I said above (and in standup) the thing I think is wrong is to say the process will "spin its own thread". Certainly a "worker" can be a process; a worker in a different implementation could also be a thread within a process. Those links are very interesting, and good for dicussion, but I don't see the thing I was objecting to addressed in either of them.

app/jobs/find_available_user_barcode_job.rb
thisisstephenbetts: There is also this, which mentions passing object IDs: https://jtway.co/its-time-for-active-job-239f7681a3ef

daryanypl: As I've said, I do not believe writing a custom serializer is appropriate for this code.  This pull request is for a README file.  Let's have a meeting and talk about the serializer option at a later time.  I can make a backlog issue for it.

daryanypl:
I've copied the conversation over to https://jira.nypl.org/browse/MLN-767 .  Let's talk about this in the next code iteration.

PR #844, opened 2021-01-20
[MLN-757] activeadmin test cases

gonuguntla:


spec/controllers/admin/books_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:book) { Book.create!(id: 1, title: "You wouldn't") }
+  let(:admin_user) { AdminUser.create!(email: 'admin@example.com', password: 'password') }
+
+  before(:each) do
+    sign_in admin_user
+  end
+
+  describe "GET index" do
+    it 'returns http success' do
+      get :index
```
thisisstephenbetts, line 17: This is all very cool, Deepika, but I don't see why you couldn't do this in a regular controller test — did that not work at all? Happy to talk more. Thanks!

gonuguntla: @thisisstephenbetts we already have regular controller test cases.  But it will not connect to active-admin .

see below image regular controller test cases already have  for book_controller.

Screen Shot 2021-01-26 at 9 42 14 AM

lib/tasks/combined_tests.rake
```diff
@@ -5,7 +5,7 @@
 # require "bundler/gem_tasks"
 require "rake/testtask"
 
-Rake::Task['test:run'].clear
+Rake::Task["test:run"].clear
```
daryanypl, line 8: I'm just curious: what was the difference here between using single and double quotes?  Was there a problem that got fixed, or is it more of a cosmetic thing?

test/functional/admin/dashboard_controller_test.rb
```diff
@@ -0,0 +1,157 @@
...
+    end
+
+    test "test index method with p_call_number_desc param" do
+      get :index, params: {order: 'p_call_number_desc'}
+      assert_response :success
+    end
+
+    test "test index method with p_call_number_asc param" do
+      get :index, params: {order: 'p_call_number_asc'}
+      assert_response :success
```
daryanypl, line 24: So these tests don't make sure that the sort is actually sorting, right?  Their purpose is to make sure the method itself is returning a success value?

test/functional/admin/teacher_sets_controller_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    test "test make_unavailable method with format js" do
+      put :make_unavailable, params: { id: @teacher_set.id }, format: :js
+      assert_response :success
+    end
+
+    test "test make_unavailable method" do
+      put :make_unavailable, params: { id: @teacher_set.id }
+      assert_response :redirect
```
daryanypl, line 39: Ok, similar question here.  It looks to me like this test and the test at line 27 are both making sure that there's _a_ redirect, but not checking where the redirect is to?

lib/tasks/combined_tests.rake
gonuguntla: No difference.  Reverted to original

test/functional/admin/teacher_sets_controller_test.rb
gonuguntla: member_action :make_unavailable, method: :put do
    teacher_set = TeacherSet.find(params[:id])
    teacher_set.availability = 'unavailable'
    teacher_set.save
    if request.format == :html
      redirect_to admin_teacher_set_path(teacher_set)
    else
      render js: "makeAvailableTeacherSet(#{teacher_set.id}, false);"
    end
  end

daryanypl (approved):

PR #845, opened 2021-01-29
update alt_email column on users to be null where it is an empty stri…

daryanypl:
…ng, and add uniqueness constraint on that column

https://jira.nypl.org/browse/MLN-715

### DIFF
--- a/db/migrate/20210129021138_set_user_alt_email_to_null_and_add_unique.rb
+++ b/db/migrate/20210129021138_set_user_alt_email_to_null_and_add_unique.rb
@@ -1,0 +1,6 @@
+class SetUserAltEmailToNull < ActiveRecord::Migration[5.0]
+  def change
+    execute "UPDATE users SET alt_email=NULL where alt_email=''"
+    add_index :users, :alt_email, name: "index_users_alt_email", unique: true
+  end
+end


db/migrate/20210129021138_set_user_alt_email_to_null_and_add_unique.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SetUserAltEmailToNullAndAddUnique < ActiveRecord::Migration[5.0]
+  def change
+    execute "UPDATE users SET alt_email=NULL where alt_email=''"
```
gonuguntla, line 5: how single quotes saved in alt_email column?

gonuguntla: In future again we will get same issue. We have to fix while saving  the alt_email.

daryanypl: Sure.  Currently, we have a unique constraint on that column.  So there can be at most only one empty alt_email row for the table.  We also have code in the users.rb model file that makes sure alt_email is unique before the user object is persisted to the database.  
I've updated the users model to take allowed_blank out for that field just now, so that should finish covering the '' scenario.

PR #850, opened 2021-02-02
Mln-702 added more barcode unit tests

daryanypl:


app/jobs/find_available_user_barcode_job.rb
```diff
@@ -38,8 +38,13 @@ def perform(user: nil, pin_code: nil, **args)
         # so that the ActiveJob mechanism would take care of retrying.
         # However, such exception handling will become available to us once
         # we upgrade our rails version.  For now, go simpler.
-        Delayed::Worker.logger.error("#{self.class.name}: user.check_barcode_uniqueness_with_sierra threw an error: #{exception.message || 'nil'}")
+        Delayed::Worker.logger.error("#{self.class.name}: user.check_barcode_uniqueness_with_sierra threw a communications error: #{exception.message || 'nil'}")
         raise exception
+      rescue ArgumentError => exception
+        # this is extremely unlikely, but shouldn't be insurmountable.
+        # let's log, and ask the following code to try again.
```
gonuguntla, line 45: @daryanypl In failure case why don't we send email to us. Since it is very important functionality.

config/application.yml
```diff
@@ -7,7 +7,7 @@ case ENV['RAILS_ENV']
 when 'local'
   ENV['MLN_INFO_SITE_HOSTNAME'] = "dev-www.mylibrarynyc.local"
   ENV['MLN_SETS_SITE_HOSTNAME'] = "dev-sets.mylibrarynyc.local"
-  ENV['DATABASE_URL'] = "postgresql://localhost/mylibnyc_local"
+  ENV['DATABASE_URL'] = "postgresql://localhost/mylibnyc_local_copy_prod"
```
gonuguntla, line 10: Please  revert to to old db name

test/unit/user_test.rb
```diff
@@ -148,13 +148,50 @@ class UserTest < ActiveSupport::TestCase
...
+    assert_equal("send_request_to_patron_creator_service called with no pin. User: #{@user.id || 0}", exception.message)
+
+    @user.pin = save_pin
+    save_barcode = @user.barcode
+    @user.barcode = nil
+    exception = assert_raise(StandardError) do
+      @user.send_request_to_patron_creator_service(nil)
+    end
+    assert_equal("send_request_to_patron_creator_service called with no barcode in user object: #{@user.id || 0}", exception.message)
+    @user.barcode = save_barcode
```
gonuguntla, line 184: is this required here?

test/unit/user_test.rb
```diff
@@ -148,13 +148,50 @@ class UserTest < ActiveSupport::TestCase
...
+    end
+    assert_equal("send_request_to_patron_creator_service called with no pin. User: #{@user.id || 0}", exception.message)
+
+    @user.pin = save_pin
+    save_barcode = @user.barcode
+    @user.barcode = nil
+    exception = assert_raise(StandardError) do
+      @user.send_request_to_patron_creator_service(nil)
+    end
+    assert_equal("send_request_to_patron_creator_service called with no barcode in user object: #{@user.id || 0}", exception.message)
```
gonuguntla, line 183: Instead of method name compare with error message

app/jobs/find_available_user_barcode_job.rb
daryanypl: Good question.  I'm relying on the code block below to retry the functionality.  This is not the best exception handling.  When we go to a higher version of rails, we'll be able to do exception handling as it's meant to be done -- there will be hooks to catch exceptions, hooks to retry_on, a mechanism for controlling how many retries and at what time intervals from the config, and a hook for what to do when all retries fail.  What you see right now is rather clumsy, and is there holding space until we can make it better.

daryanypl: Let me know if you still want an email sent.  I'm ok doing that, just not sure I should.

app/models/user.rb
```diff
@@ -135,6 +135,23 @@ def send_request_to_patron_creator_service(pin_code)
       pin_code = self.pin
     end
 
+    if pin_code.blank?
+      error_message = "send_request_to_patron_creator_service called with no pin. User: #{self.id || 0}"
```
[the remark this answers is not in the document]
daryanypl: I know, we're writing the method name twice in most of our log lines -- in the method parameter, and as part of the error message.  I know it looks clumsy, but it helps me search through the logs in CloudWatch without having to worry about phrasing the search filter query in json.

test/unit/user_test.rb
daryanypl: Same as above -- I'm writing the error message, and the method name is part of that error message.  Is this what you meant?

test/unit/user_test.rb
daryanypl: I'm not 100% sure.  The @user object is made in the setup.  I think the setup method is run once, and the @user object is reused between unit tests.  So I'm trying to clean up after myself.  I could also make a new user object directly in the test, but it takes a little bit of extra time, so I try to only do that when necessary.

app/jobs/find_available_user_barcode_job.rb
gonuguntla: Other than logs we dont have any alarm for this functionality. Better to send email

app/models/user.rb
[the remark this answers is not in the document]
gonuguntla: Ok

test/unit/user_test.rb
gonuguntla: @daryanypl  @user.barcode = save_barcode this line is not required. Because already we did comparison in above line

app/jobs/find_available_user_barcode_job.rb
gonuguntla: @daryanypl  This Pr related to test cases, Please add email notification code here.

gonuguntla (approved):

PR #857, opened 2021-02-11
Fixed cache issue

gonuguntla:


test/unit/teacher_set_test.rb
```diff
@@ -163,6 +163,23 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      resp = @teacher_set.update_subjects_via_api(subject_name_array)
+      assert_nil(resp)
+    end
+
+    # Case: 2
+    it 'update subjects' do
+      subject_ids = subjects(:subjects_one)
+      subject_name_array = ["Authors, Russiantest", "Social"]
+      resp = @teacher_set.update_subjects_via_api(subject_name_array)
+      assert_equal(subject_ids.id, resp[0])
```
thisisstephenbetts, line 179: Presumably

```      assert_equal(subjects(:subjects_two).id, resp[1])```

would also pass, right?
I'd add that as well, for completeness (and to stop me from wondering if it's true).

File: app/models/teacher_set.rb [hunk 2]
```diff
@@ -825,11 +825,14 @@ def update_included_book_list(teacher_set_record)
   def update_subjects_via_api(subject_name_array)
     LogWrapper.log('DEBUG', {'message' => 'update_subjects_via_api.start','method' => 'teacher_set.update_subjects_via_api'})
 
-    # teacher_set.rb facets_for_query uses cached results of each query
-    Rails.cache.clear unless Rails.env.test?
-
     return if subject_name_array.blank?
 
+    unless MlnConfigurationController.new.feature_flag_config('teacherset.data.from.elasticsearch.enabled')
```
thisisstephenbetts, line 830: Please put a little more comment, explaining this was necessary in the days before we had ElasticSearch. When do you think we can remove this?

thisisstephenbetts (changes requested):
Very minor changes requested, but looks good. I'm very happy to approve after that.
Thanks!

thisisstephenbetts (approved):
Looks great! Thanks Deepika

PR #858, opened 2021-02-16
Updated rails version 5.0.0.1 to 6.1.0

gonuguntla:


config/initializers/paper_trail.rb
```diff
@@ -1,3 +1,3 @@
 # frozen_string_literal: true
 
-PaperTrail.config.track_associations = false
\ No newline at end of file
+# PaperTrail.config.track_associations = false
```
thisisstephenbetts, line 3: Why is this commented out? Should we just remove this file?

test/functional/books_controller_test.rb
```diff
@@ -21,7 +21,7 @@ class BooksControllerTest < ActionController::TestCase
   end
 
   test "test book strong params" do
-    post :create, { id: 1}
+    post :create, params: { call_number: "call_number" }
```
thisisstephenbetts, line 24: Why did this change from using `id` to `call_number`?

(At some point this test could be improved by adding some more detailed assertions, but fine for now.)

thisisstephenbetts (approved):
Looks good! Nice work!

config/initializers/paper_trail.rb
gonuguntla: Ok Stephen

test/functional/books_controller_test.rb
gonuguntla: Just created book with another params instead of id.

thisisstephenbetts: Oh okay — sounds good. Thanks!

PR #859, opened 2021-02-22
[MLN-783] Changed Devise domain name to mylibrarynyc.org

gonuguntla:


config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@mylibrarynyc.org'
```
thisisstephenbetts, line 15: Let's change that to 'passwordreset@mylibrarynyc.org'. Thanks Deepika!

thisisstephenbetts (changes requested):
Thanks Deepika! I'd like the new password to be something more descriptive than "please-change-me...' I suggest using  'passwordreset@mylibrarynyc.org'. Thanks!

config/initializers/devise.rb
gonuguntla: @thisisstephenbetts Done

thisisstephenbetts (approved):
Thank you!

PR #867, opened 2021-03-02
Disable active storage for Rails 6.1.0

gonuguntla:


config/application.rb
```diff
@@ -1,8 +1,20 @@
 # frozen_string_literal: true
 
-require File.expand_path('../boot', __FILE__)
+
+require_relative 'boot'
 
 require 'rails/all'
+require "active_model/railtie"
```
thisisstephenbetts, line 7: Looks fine but why does this need to change?

db/migrate/20210226185223_create_active_storage_tables.active_storage.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
```
thisisstephenbetts, line 4: Why do we need ActiveStorage tables if we're using S3? I'd have assumed one or the other (and I'd probably go with the DB for simplicity for now, personally)

thisisstephenbetts (commented):
Got a couple of questions. Not sure why we need S3 & DB for ActiveStorage. We can chat about it on a call or on Slack if you like.

config/application.rb
thisisstephenbetts: Why go from `rails/all` I mean. I know there can be reasons, just wondering

gonuguntla: active_storage issue we can solve  in 2 ways

1) Install active storage  and add db migrations, configs

2) Disable active storage.
we do not want to load all  "rails/all".

thisisstephenbetts (approved):
Looks good — that makes sense. Thanks Deepika!

PR #871, opened 2021-03-08
Update all mln teacher-set bib ids

gonuguntla:


.rubocop_todo.yml
```diff
@@ -534,6 +534,8 @@ Metrics/AbcSize:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 432
+  Exclude:
+    - 'lib/tasks/ingest.rake'
```
gonuguntla, line 538: This file already reached maximum lines. So Excluded this file. 
 Can I create a new rake file to this functionality (update all bibs)?

thisisstephenbetts: Yes, that's cool. Thanks!

PR #872, opened 2021-03-09
[MLN-748] School object is crashing when school code is nil

gonuguntla:


app/models/school.rb
```diff
@@ -32,4 +32,23 @@ def sierra_code
...
+  # Get all active schools
+  def self.active_schools
+    active.map do |school|
+      [school.name + school_code(school), school.id] 
+    end
+  end
+
+
+  def self.inactive_school(school)
+    ["[INACTIVE] #{school.name + school_code(school)}", school.id]
```
thisisstephenbetts, line 46: This is more of a refactor than you were initially going for, but as we are repeating that `#{school.name + school_code(school)}` construction, maybe you could make a new instance method called, say, `label`, which would combine name and code, and could also add the `[INACTIVE]` label if the school isn't active.

test/functional/settings_controller_test.rb
```diff
@@ -20,4 +22,10 @@ class SettingsControllerTest < ActionController::TestCase
     assert_equal("You must be logged in to access this page", flash[:error])
     assert_response :redirect
   end
+
+  test "test index method with out school code" do
+    sign_in @user3
+    get :index, params: { settings: { contact_email: "test@gmail.com", school: { id: @school3.id } } }
+    assert_response :success
```
thisisstephenbetts, line 29: Would be great to start making slightly more detailed tests here — like checking the title of the page, looking in the HTML for the school name, for example.

test/unit/teacher_set_test.rb
```diff
@@ -177,7 +177,6 @@ class TeacherSetTest < ActiveSupport::TestCase
       subject_two = subjects(:subjects_two)
       subject_name_array = ["Authors, Russiantest", "Social"]
       resp = @teacher_set.update_subjects_via_api(subject_name_array)
-      puts ">>>>>>. #{resp} <<<<<<<"
```
thisisstephenbetts, line 180: Thanks! 😂

app/models/school.rb
```diff
@@ -32,4 +32,23 @@ def sierra_code
...
+  
+  # Get all active schools
+  def self.active_schools
+    active.map do |school|
+      [school.name + school_code(school), school.id] 
+    end
+  end
+
+
+  def self.inactive_school(school)
```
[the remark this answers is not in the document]
thisisstephenbetts: This would change if you like the suggestion below. Then there would be one instance method called, say, `label_and_id` or `label_id_array`, and that method would handle active and inactive schools appropriately.

app/controllers/settings_controller.rb
```diff
@@ -23,18 +23,17 @@ def index
...
     if current_user.present?
       @email = current_user.email
       @alt_email = current_user.alt_email
       @contact_email = current_user.contact_email
       @school = current_user.school
       @holds = current_user.holds.order("created_at DESC")
+
       #If school is inactive for current user still need to show in school drop down.
-      @schools << ["[INACTIVE] #{@school.name} (#{@school.code[1..-1].upcase})", @school.id] unless @school.active
+      @schools << School.inactive_school(@school) unless @school.active
```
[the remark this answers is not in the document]
thisisstephenbetts: But I've made some suggestions later that could further refine it.

thisisstephenbetts (commented):
Hi Deepika,
I've suggested a few changes here, to take your excellent refactors, which I really like, a bit further. These are really suggestions — please take a look and let me know what you think. I am very happy to hop on a call and discuss further, and discuss other ideas you might have.
Good work! Thanks,
Stephen.

app/models/school.rb
```diff
@@ -32,4 +32,23 @@ def sierra_code
 
     return match.sierra_code
   end
+
+  
+  # Get all active schools
+  def self.active_schools
+    active.map do |school|
+      [school.name + school_code(school), school.id]
```
thisisstephenbetts, line 40: We are repeating that #{school.name + school_code(school)} construction, maybe you could make a new instance method called, say, `label`, which would combine name and code, and could also add the [INACTIVE] label if the school isn't active. What do you think?

thisisstephenbetts: After that, we are also using the `[school.label, school.id]` construction a lot. So maybe that could go into an instance method called, say, `label_and_id`. That would be used here and in the settings controller. Please let me know what you think — just suggestions!

app/models/school.rb
```diff
@@ -32,4 +32,25 @@ def sierra_code
...
+  # Get all active schools
+  def self.active_schools_data
+    active.map do |school|
+      School.name_id(school)
+    end
+  end
+
+
+  # Parse schoool name and id
+  def self.name_id(school)
```
thisisstephenbetts, line 46: Hi Deepika, any reason that this and the next should be class methods rather than instance methods?

thisisstephenbetts: (I realise you inherited a lot of this code, so wasn't your decision initially)

app/models/school.rb
```diff
@@ -32,4 +32,23 @@ def sierra_code
...
+
+
+  # Parse schoool name and id
+  def name_id
+    status_lable = active ? "" : "[INACTIVE] "
+    ["#{status_lable}#{name + school_code(self)}", id]
+  end
+
+
+  def school_code(school)
```
thisisstephenbetts, line 50: If this is an instance method, don't need to pass `school`

app/models/school.rb
```diff
@@ -32,4 +32,23 @@ def sierra_code
 
     return match.sierra_code
   end
+
+  
+  # Get all active schools
+  def self.active_schools_data
+    active.map { |school| school.name_id }
```
thisisstephenbetts, line 39: There's a tiny shortcut you can do here. Instead of 
`active.map { |school| school.name_id }`
you can say
`active.map(&:name_id)`

test/models/school_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    @school4 = schools(:school_four)
+  end
+
+  test 'test all active schools' do
+    resp = School.active_schools_data
+    assert_equal(3, resp.count)
+  end
+
+  test 'test inactive school' do
+    resp = School.name_id(@school4)
```
thisisstephenbetts, line 19: These need to be updated for `name_id` being an instance method now

test/models/school_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    assert_equal(@school4.id, resp[1])
+  end
+
+  test 'test with school code value' do
+    resp = School.code(@school1)
+    assert_equal(" (#{@school1.code[1..-1].upcase})", resp)
+  end
+
+  test 'test with-out school code value' do
+    resp = School.code(@school3)
```
thisisstephenbetts, line 30: Same here

test/models/school_test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  test 'test inactive school' do
+    resp = School.name_id(@school4)
+    assert_equal("[INACTIVE] #{@school4.name} (#{@school4.code[1..-1].upcase})", resp[0])
+    assert_equal(@school4.id, resp[1])
+  end
+
+  test 'test with school code value' do
+    resp = School.code(@school1)
+    assert_equal(" (#{@school1.code[1..-1].upcase})", resp)
```
thisisstephenbetts, line 26: Personally I prefer it if we have the actual string to compare, rather than code like `(#{@school1.code[1..-1].upcase})`

test/models/school_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  test 'test inactive school' do
+    resp = School.name_id(@school4)
+    assert_equal("[INACTIVE] #{@school4.name} (#{@school4.code[1..-1].upcase})", resp[0])
+    assert_equal(@school4.id, resp[1])
+  end
+
+  test 'test with school code value' do
+    resp = School.code(@school1)
```
thisisstephenbetts, line 25: This needs to be updated for `code` being an instance method now

thisisstephenbetts (approved):
Looks great — thanks Deepika!

PR #886, opened 2021-03-24
Bib api code cleanup

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -145,6 +148,131 @@ def image_uri(size = :small)
...
+    # If feature flag is enabled delete data from elasticsearch.
+    if MlnConfigurationController.new.feature_flag_config('teacherset.data.from.elasticsearch.enabled')
+      # After deletion of teacherset data from db than delete teacherset doc from elastic search
+      delete_teacherset_record_from_es(teacher_set.id) if resp.destroyed?
+    end
+    teacher_set
+  end
+  
+
+  def create_or_update_teacher_set_data(req_body)
```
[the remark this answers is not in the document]
thisisstephenbetts: So you can say `TeacherSet.create_or_update_teacher_set_data` just like you'd say `TeacherSet.create`

app/models/teacher_set.rb
```diff
@@ -145,6 +148,131 @@ def image_uri(size = :small)
...
+    # If feature flag is enabled create/update data in elasticsearch.
+    if MlnConfigurationController.new.feature_flag_config('teacherset.data.from.elasticsearch.enabled')
+      # When ever there is a create/update on bib than need to create/update the data in elastic search document.
+      create_or_update_teacherset_document_in_es(TeacherSet.find(teacher_set.id))
+    end
+    teacher_set
+  end
+
+  def update_teacher_set_attribuites_from_bib_request(ts_items_info, req_body)
+    @req_body = req_body
```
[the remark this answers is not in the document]
thisisstephenbetts: Or maybe don't pass in `req_body` and just use `@req_body` below. Up to you!

app/helpers/teacher_sets_helper.rb
```diff
@@ -1,11 +1,12 @@
...
 
 module TeacherSetsHelper
+
   # Make request input body to create teacherset document in elastic search.
   # Input param ts_obj eg: <TeacherSet:0x00007fd79383a640 id: 350, title: "Step",call_number: "Teacher",
   # description: "Book", details_url: "http://catalog.nypl.org/record=b21378444~S1","updated-7571-Step up to the plate, Maria Singh">
   # Expected response from this method {:title=>"Step Up", :description=> "Book", :contents=>"Step Up", :id=>350}
-
+  
   def teacher_set_info(ts_obj)
```
thisisstephenbetts, line 10: Maybe should be an instance method?

app/controllers/concerns/bib_service.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module BibService
```
thisisstephenbetts, line 3: Maybe call `bib_service_helper.rb` and move alongside other helper modules?

thisisstephenbetts (changes requested):
A few minor changes — great work so far! Thank you!

test/unit/helpers/teacher_sets_helper_test.rb
```diff
@@ -2,88 +2,74 @@
...
+      self.instance_variable_set(:@req_body, SIERRA_USER["data"][0])
+      resp = all_var_fields('300')
+      assert_equal(["physical desc"], resp)
+    end
+  end
+
+  describe 'test grade value method' do
+    it 'test pre-k grades' do
+      resp = grade_val('K')
+      assert_equal(0, resp)
```
thisisstephenbetts, line 58: I am fine with writing this as 
```
assert_equal 0, grade_val('K')
```

app/models/teacher_set.rb
```diff
@@ -754,22 +893,20 @@ def update_teacher_set_set_type_value(set_type_val)
   # If subfields.content type is "Book Club Set" set_type value  stored as 'single' in teacher_sets table.
   # case 2: If it is not present in subfields.content, derive the set_type from the number of distinct books attached to a TeacherSet.
   # If teacher-set-books exactly 1, it's a Bookclub Set; else it's a Topic Set.
-  def update_set_type(set_type_val)
+  def self.get_set_type_value(t_set, set_type_val)
```
thisisstephenbetts, line 896: I think this would be clearer as an instance method, because the way it's called, it's looking for a 526 field, and only if it doesn't find it does this logic come in. So that would be, in my opinion, better encapsulated within the method.

I would call this
```
def derive_set_type
  set_type = self.var_field_data('526')

  return set_type unless set_type.nil?

  #  then go into the fallback calculation looking at #books
```

thisisstephenbetts: Just an idea, something like that. See what you think!

app/helpers/teacher_sets_helper.rb
thisisstephenbetts: My mistake! Probably don't need to pass in `ts_obj` then and can just use `self` in the method?

app/helpers/teacher_sets_es_helper.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module TeacherSetsEsHelper
+  # Make teacherset object from elastic search json
+  def create_ts_object_from_es_json(json)
+    arr = []
```
thisisstephenbetts, line 6: Maybe give this variable a more descriptive name — like `teacher_sets`  or `teacher_sets_from_es`

app/helpers/bibs_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+    if physical_description.blank? && validate
+      req_body_params << { value: physical_description, error_msg: PHYSICAL_DESCRIPTION_EMPTY[:msg]}
+    end
+
+    validate_empty_values(req_body_params) if req_body_params.present?
+  end
+
+
+  def var_field(marc_tag, req_body)
+    @req_body = req_body
```
thisisstephenbetts, line 32: This line is a bit confusing.

It maybe easier to have `var_field_data` as a TeacherSet method, which calls `var_field` with the req_body.

So you'd have TeacherSetHelper#var_field_data like

```
def var_field_data(marc_tag, merge)
  var_field(@req_body, marc_tag, merge)
end
```

And try to avoid setting `@req_body` in _any_ helper methods

thisisstephenbetts: Then the implementation of `var_field` would basically be the current implementation of `var_field_data` — it's flipping how those two methods work: 

```
def var_field(req_body, marc_tag, merge = true)
    if merge == true
      req_body['varFields'].detect { |hash| hash['marcTag'] == marc_tag }['subfields'].map { |x| x['content'] }.join(', ')
    else
      req_body['varFields'].detect { |hash| hash['marcTag'] == marc_tag }['subfields'].detect { |hash| hash['tag'] == 'a' }['content']
    end
  rescue StandardError
    nil
  end
```

Then line 11 wouldn't change, it just wouldn't set the `@req_body`

app/models/teacher_set.rb
```diff
@@ -741,10 +905,25 @@ def get_set_type(set_type)
...
                             'method' => 'teacher_set.update_teacher_set_set_type_value'})
+  rescue StandardError => e
+    raise DBException.new(TEACHERSET_SETTYPE_ERROR[:code], TEACHERSET_SETTYPE_ERROR[:msg])
+  end
+
+
+  def update_set_type(set_type_val)
+    begin
+      set_type = derive_set_type(set_type_val)
+      LogWrapper.log('INFO', {'message' => "Teacher set set_type value: #{set_type}",'method' => 'teacher_set.update_set_type'})
```
thisisstephenbetts, line 918: If you don't want this log message here, I would get rid of it.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,298 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+
+  describe 'update set_type value in teacher_set table' do
+    it 'test update_set_type value method' do
+      resp = nil
+      set_type_val = @teacher_set3.set_type
+      @mintest_mock1.expect(:call, set_type_val, [set_type_val])
+      @mintest_mock2.expect(:call, true, [set_type: set_type_val])
+
+      @teacher_set3.stub :derive_set_type, @mintest_mock1 do
+        TeacherSet.stub :update, @mintest_mock2 do
```
thisisstephenbetts, line 405: As `update` is an instance method, I think this is clearer if you stub the instance, not the class.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,299 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      end
+      assert_equal(set_type_val, resp)
+      @mintest_mock1.verify
+      @mintest_mock2.verify
+    end
+
+    it 'test StandardError while updating the set_type' do
+      resp = nil
+      teacher_set = TeacherSet.new
+      invalid_set_type = "&&&&&&7"
```
thisisstephenbetts, line 417: This is confusing for me, because it looks like the error is being raised because the `set_type` is invalid, but it's actually because `update` is being called with the wrong parameters. Probably better to just have a normal or nil `set_type` so it doesn't look like that is the problem.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
     end
   end
 
+  describe 'create_or_update_teacherset ' do
+    it 'test create or update teacherset' do
+      ts_items_info = {bibs_resp: SIERRA_USER, total_count: 1, available_count: 1, availability_string: 'available'}
+      TeacherSet.instance_variable_set(:@req_body, SIERRA_USER["data"][0])
```
thisisstephenbetts, line 199: So I think this is setting @req_body on the Class! You probably want to put this on line 202 and call it on `teacher_set`

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      bnumber = {bnumber: "b7899158"}
+      teacher_set = TeacherSet.new(bnumber: "b7899158")
+      subjects_arr = ["Elections."]
+
+      resp = nil
+      @mintest_mock2.expect(:call, ts_items_info, [7899158])
```
thisisstephenbetts, line 205: So. I realise this may be a bit annoying for you Deepika — and I honestly appreciate all the work you've put in to this — but I've come to the conclusion that this is probably too complex for a unit test. 

For a unit test we typically want to test the outputs with different inputs and _not_ worry about what happens within the method. I think where we _do_ want to use a mock there is if we're calling an external service (like an API). So `get_items_info_from_bibs_service`, `update_subjects_via_api` should be mocked, but that may be it.

That way you can just set up the ts_item_info, and then check the attributes of the TeacherSet instance created. 

Does that make sense? Sorry if this is causing a lot of extra work.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+    it 'test delete_teacher_set_method' do
+      bib_id = '1234'
+      teacher_set = TeacherSet.new(bnumber: "b#{bib_id}")
+      @mintest_mock1.expect(:call, teacher_set, [bib_id])
+  
+
+      resp = nil
+      TeacherSet.stub :get_teacher_set_by_bnumber, @mintest_mock1 do
+        teacher_set.stub :destroy, teacher_set do
+          teacher_set.stub :delete_teacherset_record_from_es, teacher_set do
```
thisisstephenbetts, line 259: So for this one, I think `delete_teacherset_record_from_es` should be mocked, but I don't think anything else need be.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+    end
+
+    it 'test bib record not found exception' do
+      bib_id = '1234'
+      resp = nil
+      teacher_set = TeacherSet.new(bnumber: "")
+      @mintest_mock1.expect(:call, [], [bib_id])
+      resp = assert_raises(BibRecordNotFoundException) do
+        TeacherSet.stub :get_teacher_set_by_bnumber, @mintest_mock1 do
+          teacher_set.stub :delete_teacherset_record_from_es, teacher_set do
```
thisisstephenbetts, line 275: Does this need to be stubbed? It doesn't look like it should be called according to https://github.com/NYPL/MyLibraryNYCApp/blob/044b9dfdb51f76aae2971dd2b5f885752bc83edf/app/models/teacher_set.rb#L244

thisisstephenbetts: It feels like we can test `assert_raises(BibRecordNotFoundException)` on a nonexistent bib_id without mocking or stubbing anything.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+       :created_at => "2014-10-19T01:00:00+0000",:updated_at => "2014-10-19T01:00:00+0000",:available_copies => 1,
+       :bnumber => "B456",:set_type => "set_type",:area_of_study => "area_of_study",:subjects => []}
+    end
+
+    it 'update teacher-set document in es' do
+      resp = nil
+      @mintest_mock1.expect(:call, @expected_resp)
+      @mintest_mock2.expect(:update_document_by_id, @es_doc, [@expected_resp[:id], @expected_resp])
+      
+      @ts_obj.stub :teacher_set_info, @mintest_mock1 do
```
thisisstephenbetts, line 323: Similarly, I don't think we need stub `teacher_set_info` — once you've created @ts_obj you should be able to just call it. It does make sense to mock `update_document_by_id` on ElasticSearch though — that looks good!

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+          resp = @ts_obj.create_or_update_teacherset_document_in_es
+        end
+      end
+
+      @mintest_mock1.verify
+      @mintest_mock2.verify
+      assert_equal(188, resp)
+    end
+
+    it 'test StandardError while updating elastic-search' do
```
thisisstephenbetts, line 334: Hmmm, I don't understand this one. Why is StandardError raised?

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+    it 'test delete_teacherset_record_from_es method' do
+      expected_resp = {"_index" => "teacherset", "_type" => "teacherset", "_id" => @teacher_set.id, "_version" => 2, "result" => "deleted", 
+        "_shards" => {"total" => 2, "successful" => 1, "failed" => 0}, "_seq_no" => 294, "_primary_term" => 2}
+      @mintest_mock1.expect(:delete_document_by_id, expected_resp, [@teacher_set.id])
+      resp = nil
+      ElasticSearch.stub :new, @mintest_mock1 do
+        resp = @teacher_set.delete_teacherset_record_from_es
+      end
+      @mintest_mock1.verify
+      assert_equal(184, resp)
```
thisisstephenbetts, line 364: I'm a bit confused here — what is `184` here? Where does that come from?

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+    end
+
+    it 'test StandardError while deleting elastic-search document' do
+      expected_resp = {"_index" => "teacherset", "_type" => "teacherset", "_id" => @teacher_set.id, "_version" => 2, "result" => "deleted", 
+        "_shards" => {"total" => 2, "successful" => 1, "failed" => 0}, "_seq_no" => 294, "_primary_term" => 2}
+      @mintest_mock1.expect(:delete_document_by_id, expected_resp, [nil])
+      resp = nil
```
thisisstephenbetts, line 371: This line doesn't do anything, right?

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      resp = assert_raises(StandardError) do
+        ElasticSearch.stub :new, @mintest_mock1 do
+          resp = @teacher_set.delete_teacherset_record_from_es
```
thisisstephenbetts, line 374: Why are we assigning to `resp` on this line?

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      @mintest_mock1.expect(:delete_document_by_id, expected_resp, [@teacher_set.id])
+      resp = nil
+      ElasticSearch.stub :new, @mintest_mock1 do
+        resp = @teacher_set.delete_teacherset_record_from_es
+      end
+      @mintest_mock1.verify
+      assert_equal(184, resp)
+    end
+
+    it 'test StandardError while deleting elastic-search document' do
```
thisisstephenbetts, line 367: This is a good test — basically it's saying that if we get a `nil` response from ES then an error is thrown, right? 

If so would be good to describe that a bit more, something like "TeacherSet#delete_teacherset_record_from_es raises StandardError if ES doesn't return the expected response"

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+
+
+  describe 'update set_type value in teacher_set table' do
+    it 'test update_set_type value method' do
+      resp = nil
+      set_type_val = @teacher_set3.set_type
+      @mintest_mock1.expect(:call, set_type_val, [set_type_val])
+      @mintest_mock2.expect(:call, true, [set_type: set_type_val])
+
+      @teacher_set3.stub :derive_set_type, @mintest_mock1 do
```
thisisstephenbetts, line 401: So, for example, here I don't think we _need_ to know that `derive_set_type` was called in a unit test. We only need to know that for given inputs, the right outcomes happened (i.e., the `set_type` was updated. I think that's all that's needed for a unit test.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+        @teacher_set3.stub :update, @mintest_mock2 do
+          resp = @teacher_set3.update_set_type(set_type_val)
+        end
+      end
+      assert_equal(set_type_val, resp)
+      @mintest_mock1.verify
+      @mintest_mock2.verify
+    end
+
+    it 'test StandardError while updating the set_type' do
```
thisisstephenbetts, line 411: Please add a bit more detail here — why is StandardError being raised?

gonuguntla: @thisisstephenbetts Yes Stephen I read all comments. I mocked  and stubbed each and everything inside the method also. Now I will fix it as per you comments.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+        end
+      end
+      assert_equal(TEACHERSET_SETTYPE_ERROR[:code], resp.code)
+      assert_equal(TEACHERSET_SETTYPE_ERROR[:msg], resp.message)
+    end
+  end
+
+
+  describe 'get set-type value' do
+    it 'test book club set set_type value' do
```
thisisstephenbetts, line 429: I don't think we use `get_set_type` anywhere anymore — should delete the method and this test and the next one. (It is used in one migration, but that should probably be updated to make it a 'no-op' migration.)

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      resp = @teacher_set.get_set_type('single')
+      assert_equal(TeacherSet::BOOK_CLUB_SET, resp)
+    end
+
+    it 'test topic set set_type value' do
+      resp = @teacher_set.get_set_type('multi')
+      assert_equal(TeacherSet::TOPIC_SET, resp)
+    end
+
+    it 'get set_type value from bib_response' do
```
thisisstephenbetts, line 439: This looks good! Thanks

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      bib_id = 7899158
+      resp = OpenStruct.new(data: SIERRA_USER["data"], code: 200)
+      @mintest_mock1.expect(:call, resp, [bib_id])
```
thisisstephenbetts, line 442: If we're not verifying the mock, then I think we only need to stub out the response.
Maybe something like 

```
@teacher_set3.stub :send_request_to_bibs_microservice, resp do
```

I _think_ that should work...? If we don't care how it's called and are not running `verify` then I think it's clearer to do without the mock if we can. I'm basing this off https://www.rubydoc.info/gems/minitest/4.2.0/Object:stub, so may well be wrong.

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +193,293 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      end
+      assert_equal(TeacherSet::TOPIC_SET, resp)
+    end
+  end
+
+
+  describe 'clear primary_subject column in teacher_set table' do
+    it 'test clean_primary_subject method' do
+      resp = @teacher_set4.clean_primary_subject
+      assert_equal(144, resp)
```
thisisstephenbetts, line 455: What is `144` here?

thisisstephenbetts: This test doesn't seem to be testing anything about `clean_primary_subject` — length, whitespace, trailing-period, empty string. Probably better to either test each of those, or else just remove this test.

app/models/teacher_set.rb
```diff
@@ -145,6 +149,166 @@ def image_uri(size = :small)
...
+    end
+    teacher_set
+  end
+
+
+  # When Delete request comes from sierra, than delete teacher set document in elastic search.
+  def delete_teacherset_record_from_es
+    ts_id = self.id
+    resp = ElasticSearch.new.delete_document_by_id(ts_id)
+    return unless resp["result"] == "deleted"
```
thisisstephenbetts, line 268: Why do we return here? Would that ever happen?

test/unit/teacher_set_test.rb
```diff
@@ -182,13 +186,169 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      assert_equal(@teacher_set.bnumber, resp[:bnumber])
+      assert_equal(@teacher_set.subjects.count, resp[:subjects].count)
+    end
   end
 
 
+  describe 'update set_type value in teacher_set table' do
+    it 'test update_set_type value method' do
+      resp = nil
+      set_type_val = @teacher_set3.set_type
```
thisisstephenbetts, line 294: I think this would look more meaningful if you changed the `set_type`. Maybe assert what the `@teacher_set3.set_type` equals initially; run the update methods to something else; assert it equals the new value after that.

test/unit/teacher_set_test.rb
```diff
@@ -182,12 +186,218 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      assert_equal(ELASTIC_SEARCH_STANDARD_EXCEPTION[:msg], resp.message)
+      elasticsearch_adapter_mock.verify
+    end
+  end
+
+
+  # Delete teacher-set document in ES
+  describe "Delete teacherset record from elastic search" do
+
+    it 'Delete teacher_set document from Elastic search' do
```
thisisstephenbetts, line 288: Nice! this looks good.

test/unit/teacher_set_test.rb
```diff
@@ -182,12 +186,218 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      end
+      assert_equal(ELASTIC_SEARCH_STANDARD_EXCEPTION[:code], resp.code)
+      assert_equal(ELASTIC_SEARCH_STANDARD_EXCEPTION[:msg], resp.message)
+      elasticsearch_adapter_mock.verify
+    end
+  end
+
+
+  # Delete teacher-set document in ES
+  describe "Delete teacherset record from elastic search" do
```
thisisstephenbetts, line 286: So a good practice for these descriptions is to say what method is being tested in the `describe`, and then say what you are specifically testing in the `it` line. 

For example:

```
  describe "TeacherSet#delete_teacherset_record_from_es" do
      it "should call ElasticSearch#delete_document_by_id"
          # test here
```

test/unit/teacher_set_test.rb
```diff
@@ -182,12 +186,218 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      resp = nil
+      ElasticSearch.stub :new, @elasticsearch_adapter_mock do
+        @teacher_set.delete_teacherset_record_from_es
+      end
+      @elasticsearch_adapter_mock.verify
+    end
+
+    it 'raises ElasticsearchException when ElasticSearch raises NotFound' do
+      detail_msg = "Teacher set data not found."
+      @elasticsearch_adapter_mock.expect(:delete_document_by_id, nil) do
```
thisisstephenbetts, line 300: I really like the named mocks! That's very helpful

test/unit/teacher_set_test.rb
```diff
@@ -182,12 +186,218 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      es_resp = {"found" => true, "result" => 'deleted'}
+      bib_id = '999999'
+      # Create dedicated teacher-set record with bib_id '99999'.
+      TeacherSet.new(bnumber: "b#{bib_id}").save
+      resp = nil
+      # Delete teacher_set record
+      TeacherSet.stub_any_instance :delete_teacherset_record_from_es, es_resp do
+        resp = TeacherSet.delete_teacher_set(bib_id)
+      end
+      assert_equal("b#{bib_id}", resp.bnumber)
```
thisisstephenbetts, line 201: This is great — to be really complete you could show that there is no TeacherSet with that ID anymore.

test/unit/teacher_set_test.rb
```diff
@@ -182,12 +186,218 @@ class TeacherSetTest < ActiveSupport::TestCase
...
+      bibd_id = SIERRA_USER["data"][0]['id']
+
+      ts_items_info = {bibs_resp: SIERRA_USER, total_count: 1, available_count: 1, availability_string: 'available'}
+      
+      TeacherSet.stub_any_instance :get_items_info_from_bibs_service, ts_items_info do
+        TeacherSet.stub_any_instance :create_or_update_teacherset_document_in_es, es_resp do
+          resp = TeacherSet.create_or_update_teacher_set(SIERRA_USER["data"][0])
+        end
+      end
+      assert_equal(@teacher_set6.bnumber, resp.bnumber)
```
thisisstephenbetts, line 235: Maybe use `bib_id` here too.

thisisstephenbetts (approved):
Great work! Thanks Deepika

PR #922, opened 2021-04-27
Removed unwanted logs

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -1084,8 +1081,6 @@ def update_notes(teacher_set_notes_string)
   # Updated MLN DB with Total copies and available copies.
   def update_available_and_total_count(bibid)
     response = get_items_info_from_bibs_service(bibid)
-    LogWrapper.log('INFO','message' => "TeacherSet available_count: #{response[:available_count]}, total_count: #{response[:total_count]},
-    availability: #{response[:availability_string]}", b_number: "#{bibid}")
```
gonuguntla, line 1088: available_count, total_count already logging in anther method. So removing this log

app/controllers/teacher_sets_controller.rb
```diff
@@ -11,47 +11,44 @@ class TeacherSetsController < ApplicationController
...
-      # Determine what facets are selected based on query string
-      @facets.each do |f|
-        f[:items].each do |v|
-          k = f[:label].underscore
-          v[:selected] = params.keys.include?(k) && params[k].include?(v[:value].to_s)
-        end
+    # Feature flag: 'teacherset.data.from.elasticsearch.enabled = true' means gets teacher-set documents from elastic search.
+    # teacherset.data.from.elasticsearch.enabled = false means gets teacher-set data from database.
+    if MlnConfigurationController.new.feature_flag_config('teacherset.data.from.elasticsearch.enabled')
+      LogWrapper.log('INFO', {'message' => "Calling elastic search to get teacher-sets",
```
thisisstephenbetts, line 17: This is a nice refactor! Thanks

Do we need log statements and the one on line 24? I'm fine if they're useful but if not let's drop them.

thisisstephenbetts (approved):
There are a couple more `INFO` statements which you could remove, but either way, I'm fine with approving this. Thank you!

PR #931, opened 2021-05-10
Updated availability in teacher-set table

gonuguntla:


app/models/teacher_set.rb
```diff
@@ -102,6 +102,41 @@ def holds_for_user(user, hold_id)
...
   end
 
+  
+  def teacher_set_availability(available_copies)
+    available_copies.to_i > 0 ? AVAILABLE : UNAVAILABLE
+  end
+
+
+  def calculate_available_copies(status, quantity, current_user, hold_id)
+    if status == 'cancelled'
```
thisisstephenbetts, line 112: Would be worth adding a comment here saying that `quantity` is not set for cancellations, so must be calculated

test/unit/teacher_set_test.rb
```diff
@@ -399,6 +399,63 @@ def test_teacher_set_query
...
+
+  describe 'teacher_set#calculate_available_copies' do
+    it 'calculate teacher-set available copies while cancelling the hold' do
+      available_copies, availability = @teacher_set2.calculate_available_copies('cancelled', nil, @user, @hold1.id)
+      assert_equal(2, available_copies)
+      assert_equal('available', availability)
+    end
+
+    it 'calculate teacher-set available copies while creating the hold' do
+      available_copies, availability = @teacher_set2.calculate_available_copies('create', @hold1.quantity)
```
thisisstephenbetts, line 424: Would be useful to show how many available_copies there are before creating the hold

test/unit/teacher_set_test.rb
```diff
@@ -399,6 +399,63 @@ def test_teacher_set_query
...
+    it 'update teacher-setavailability while cancelling the hold ' do
+      resp = @teacher_set2.update_teacher_set_availability_in_db('cancelled', nil, @user, @hold1.id)
+      assert_equal(true, resp)
+    end
+  end
+
+
+  describe 'teacher_set#calculate_available_copies' do
+    it 'calculate teacher-set available copies while cancelling the hold' do
+      available_copies, availability = @teacher_set2.calculate_available_copies('cancelled', nil, @user, @hold1.id)
```
thisisstephenbetts, line 418: Would be useful to show how many available_copies there are before cancelling the hold

thisisstephenbetts: (i.e., do an assertion on the number of copies)

test/unit/teacher_set_test.rb
```diff
@@ -399,6 +399,63 @@ def test_teacher_set_query
...
+      end
+      assert_equal(resp['result'], "updated")
+      elasticsearch_adapter_mock.verify
+    end
+  end
+
+
+  describe 'teacher_set#teacher_set_availability' do
+    it 'get teacher_set availability based on the available copies' do
+      ts_availability = @teacher_set2.teacher_set_availability(@teacher_set.available_copies)
```
thisisstephenbetts, line 454: Hmmm, I don't understand what's going on here. Why does the `teacher_set_availability` method take an argument? I guess because it was moved from a different class/module. Looks to me like it would be much better and clearer if `teacher_set_availability` looked at the `available_copies` within the implementation of the instance method.

You can see the problem here, in that you're looking at availability in `@teacher_set2` based on the available copies of `@teacher_set` so the answer isn't going to be meaningful.

app/models/teacher_set.rb
```diff
@@ -102,6 +102,41 @@ def holds_for_user(user, hold_id)
...
+      available_copies = self.available_copies - quantity.to_i
+    end
+    available_copies
+  end
+
+
+  def update_teacher_set_availability_in_db(status, quantity=nil, current_user=nil, hold_id=nil)
+    available_copies, availability = calculate_available_copies(status, quantity, current_user, hold_id)
+    self.available_copies = available_copies
+    self.availability = teacher_set_availability
```
thisisstephenbetts, line 126: Based on your last comment at planning — about the need to think through implications of changing `availability_count` etc., should `availability` be saved to the DB? Could we just remove the column from the TeacherSet table, and rename the `teacher_set_availability` method to `availability`? Happy to discuss further — let me know what you think!

thisisstephenbetts: That way there would be no way for the `availability` column and `available_copies` to be inconsistent.

test/functional/holds_controller_test.rb
```diff
@@ -24,56 +28,88 @@ class HoldsControllerTest < ActionController::TestCase
...
   test "should cancel hold" do
-    post :cancel, params: { id: @hold1.access_key }
+    post :cancel, params: { id: @hold2.access_key }
     assert_response :success
   end
 
   test "update teacher-set available_copies and cancel hold successfully" do
     sign_in @user
-    expected_ts_available_copies = 4
+    expected_ts_available_copies = 3
```
thisisstephenbetts, line 42: I would just write `3` in line 59, not assign to a variable. Just my personal preference.

test/functional/holds_controller_test.rb
```diff
@@ -24,56 +28,88 @@ class HoldsControllerTest < ActionController::TestCase
...
+              "_id" => @hold4.teacher_set.id, "_version" => 11, "result" => "updated", 
+              "_shards" => {"total" => 0, "successful" => 1, "failed" => 0}}
+
+    TeacherSet.stub_any_instance :update_teacher_set_availability_in_elastic_search, es_doc do
+      post :update, params: { id: @hold4.access_key, hold_change: {"comment" => "qqq", "status" => "cancelled"}, hold: {status: "MyText"} }
+    end
     assert_nil(flash[:notice])
   end
 
   test "Should fail hold cancelaltion" do
```
thisisstephenbetts, line 76: I think this test is fine, but this string doesn't describe whats happening. Basically, it's showing that the available copies doesn't change if the hold status is _not_ "cancelled", so please write the description string like that. Thanks!

test/functional/holds_controller_test.rb
```diff
@@ -24,56 +28,88 @@ class HoldsControllerTest < ActionController::TestCase
...
-    assert_equal(resp_hold_obj.status, "new")
     # teacher-set available_copies also not changed, because hold is not cancelled.
+    assert_equal(resp_hold_obj.status, "new")
     assert_equal(expected_ts_available_copies, TeacherSet.find(resp_hold_obj.teacher_set_id).available_copies)
     assert_response :redirect
   end
 
   test "create hold and update teacher-set available_copies" do
-    expected_ts_available_copies = 0
+    expected_ts_available_copies = 1
```
thisisstephenbetts, line 99: Please don't use `expected_ts_available_copies` here — just write `1` in the `assert_equals` statement below.

test/test_helper.rb
```diff
@@ -10,7 +10,7 @@
   add_filter 'lib/tasks/sync_users.rake'
 end
 # fail unit tests if total coverage dips below acceptable limit
-SimpleCov.minimum_coverage 62
+SimpleCov.minimum_coverage 63
```
thisisstephenbetts, line 13: Awesome!!

test/unit/teacher_set_test.rb
```diff
@@ -399,6 +389,69 @@ def test_teacher_set_query
     end
   end
 
+
+  describe 'teacher_set#update_teacher_set_availability_in_db' do
+    it 'update teacher-setavailability while creating the hold ' do
```
thisisstephenbetts, line 394: For this test, and that on line 399, this should really assert the value of `availabile_copies` and `availability` before the test, and then do it again afterwards. As it is, all this really tests is that an error isn't thrown, not that it is doing what it said it would.

thisisstephenbetts: You do this well in the tests on line 406 and 416 — something like that would be good.

thisisstephenbetts (changes requested):
Hi Deepika,
thanks very much — I'm really keen to keep improving our tests, and making them as clear as possible so I have a couple more suggestions. Hopefully should be quick to fix. Thanks!
Stephen.

thisisstephenbetts (approved):
Great work — thanks!

PR #935, opened 2021-05-20
Uploading MLN calendar from google document

gonuguntla:


app/admin/documents.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
```
gonuguntla, line 1: Still working on code comments/document and test cases

thisisstephenbetts (changes requested):
Couple of minor changes. Looks good though. Obviously would like to see a demo and will review again once all tests are complete. Thanks!

app/admin/documents.rb
```diff
@@ -0,0 +1,102 @@
...
+ActiveAdmin.register Document do
+  menu :priority => 12
+
+  permit_params :file, :event_type, :file_name, :url
+
+  controller do
+    def create
+      attrs = params[:document]
+      @document = Document.new
+      @document[:event_type] = attrs[:event_type]
```
thisisstephenbetts, line 12: Lines 12-18 _could_ be factored into a method so it can be reused in the update method, but not a big deal, just something to consider for the future.

app/models/document.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def validate_event_type
+    errors.add(:event_type, 'Please select event_type') if event_type.to_s == "0"
+  end
+
+
+  def event_type_already_exist
+    return unless Document.where(event_type: event_type).present?
+
+    errors.add(:event_type, "#{event_type.titleize} type already created. Please use another type.")
```
thisisstephenbetts, line 24: Please update the 2nd sentence to say "Please use another type, or use the 'edit' link if you are trying to update "#{event_type.titleize}"

app/models/document.rb
```diff
@@ -0,0 +1,56 @@
...
+    # Split google document url than collect document-id.
+    # document_id = 1iBzIYM_GG5OCXkuF4vKwSYRFaH3gd8Q_kuDrqT7Iu4U
+    document_id = URI.split(url)[5].split('/')[3]
+    GoogleApiClient.export_file(document_id, "application/pdf")
+  rescue StandardError => error
+    errors.add(:url, google_client_error_message(error))
+  end
+
+
+  def self.calendar_event
```
thisisstephenbetts, line 40: It could be confusing for future developers to have both `calendar_event` and `calendar_of_event` — they sound the same. Maybe we could just have the scope `calendar_of_event` just return the `first` and then get rid of this method. Would that work?

app/views/home/mln_calendar.html.erb
```diff
@@ -0,0 +1,10 @@
+<% unless @calendar_event.present? %>
+  <div role="main">
+   <h2 class="pg--title">Mylibrarynyc calendar</h2>
+    <div class="pg--content">
+      <table>
+        Mylibrarynyc calendar not found. Please create calendar.
```
thisisstephenbetts, line 6: If this is the error that regular patrons see we should not say "Please create calendar." — you can just remove that second sentence.

test/functional/admin/documents_controller_test.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'minitest/stub_any_instance'
+
+module Admin
+  class DocumentsControllerTest < ActionController::TestCase
```
thisisstephenbetts, line 7: So these tests are just testing responses. They're not testing that what we want to happen has happened. Let's discuss at our meeting today, but ahead of that meeting please think about, for each of these methods, what we want the method to do and how we could test them.

thisisstephenbetts (changes requested):
I'm requesting a few very minor changes. Hopefully we can get these sorted out after our meeting this afternoon. Thanks!

app/models/document.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+class Document < ActiveRecord::Base
+
+  validate :validate_event_type, :on => [:create, :update]
+  validate :event_type_already_exist, :on => :create
+
+  EVENTS = [['-- Select --', 0], ['Calendar of events', 'calendar_of_event']].freeze
```
thisisstephenbetts, line 8: Y'know, it would be better if we changed `calendar_of_event` to `calendar_of_events` with an s on the end. Thanks!

thisisstephenbetts (approved):
Looks good — thanks!

PR #950, opened 2021-05-27
Added feature flag for pin-password

gonuguntla:


app/models/user.rb
```diff
@@ -209,6 +216,9 @@ def validate_pin_pattern
     if pin && pin&.scan(/(.)\1{2,}/)&.empty? && pin.scan(/(..)\1{1,}/)&.empty? == true
       true
     else
+      msg = 'PIN does not meet our requirements. Please try again.'
+      return errors.add(:pin, msg) unless MlnConfigurationController.new.feature_flag_config('signup.pin_password.enabled')
```
thisisstephenbetts, line 220: Should this return `false`?

app/views/devise/registrations/new.html.erb
```diff
@@ -59,16 +59,30 @@
   </div>
 
   <div class="field-group">
-    <%= label_tag :user_pin, "PIN/PASSWORD", :id => 'userPin-label' %>
+    <% pin_password_flag  = MlnConfigurationController.new.feature_flag_config('signup.pin_password.enabled') %>
+    <% pin = pin_password_flag ? "PIN/PASSWORD" : "PIN" %>
```
thisisstephenbetts, line 63: Would probably call this `pin_label` rather than `pin`, but no big deal

app/models/user.rb
thisisstephenbetts: (So may need to be a conditional block, not a one-line)

gonuguntla, line 220: I written condition block, Rubocop complained about it. After that i written like this

thisisstephenbetts (approved):
Looks good — thanks!

PR #1188, opened 2023-03-15
[SEB-2614] dockerize app and deploy to ECS

shuynh:
https://jira.nypl.org/browse/SEB-1847

PR #1195, opened 2023-03-23
[SEB-1847] dockerize app and deploy to ECS

shuynh:
https://jira.nypl.org/browse/SEB-1847

formerly https://github.com/NYPL/MyLibraryNYCApp/pull/1188

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,39 +21,29 @@ jobs:
           export ENV="production";
         fi
       env:
-      - LOCAL_TAG_NAME="${ENV}-$TRAVIS_BUILD_NUMBER"
       - ECR_URL=946183545209.dkr.ecr.us-east-1.amazonaws.com/my-library-nyc-app
+      - AWS_DEFAULT_REGION=us-east-1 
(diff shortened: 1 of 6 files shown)


Dockerfile.multistage
```diff
@@ -0,0 +1,54 @@
...
+      bundle install --jobs 30
+
+## webpack and app for now
+# TODO: eventually split up the webpack part
+FROM builder AS app
+COPY --from=bundler /usr/local/bundle /usr/local/bundle
+
+# set env vars
+ENV APP_HOME /home/app/MyLibraryNYCApp
+ENV RAILS_ENV=qa
```
gonuguntla: RAILS_ENV=qa - It should be Environment variable

.travis.yml
```diff
@@ -1,4 +1,4 @@
-language: ruby
+language: minimal
```
gonuguntla, line 1: Why you changed language name?

shuynh: since we're using docker now we don't need Travis to install ruby

PR #1352, opened 2024-02-13
Added condition for expiration date

gonuguntla:


app/models/user.rb
```diff
@@ -285,7 +285,7 @@ def barcode_available_in_sierra?
 
   def calculate_next_recurring_event_date(current_date=Date.today, month=6, day=30)
     # Check if it's after June 30th or on June 30th
-    future_date = if current_date.month > month || (current_date.month == month && current_date.day >= day)
+    future_date = if current_date.month >= month || (current_date.month == month && current_date.day >= day)
```
gonuguntla, line 288: @nonword If your ok with this condition approve otherwise revoke it.

nonword: It seems that the new policy is: use June 30 of next year if the current date is >= June 1. Is that right? If so, I believe the second part of the clause is redundant. And the comment should probably be updated to clarify intent.

gonuguntla:
After discussing with sierra team, this code is not required

PR #1366, opened 2024-04-05
Upgraded react 17 to react 18

gonuguntla:


.travis.yml
```diff
@@ -5,15 +5,15 @@ services:
 
 branches:
   only:
-  - qa
+  - MLN-1252-react18
```
gonuguntla: This code need to revert before merging the code

PR #1385, opened 2024-06-06
Implemented typography

gonuguntla:


app/javascript/components/Home/Home.jsx
```diff
@@ -92,6 +92,7 @@ export default function Home(props) {
                 id="mln-campaign-hero"
                 text="Welcome to MyLibraryNYC"
                 color="ui.white"
+                className="noTextTransform"
```
gonuguntla: I added noTextTransform class here.

bigfishdesign13: Why not remove the following `h1` style from the stylesheet?
```
h1 {
    text-transform: capitalize !important;
}
```

bigfishdesign13 (commented):
Overall, this looks good.  I just had one quetion.

Also, I see there was one line from the spreadsheet that was not addressed.  Also, I realized that I did not include changes for the `Cancel order` page, so I added a few more lines to the spreadsheet.

Shouldn't there be more than seven pages included in this PR?

app/javascript/components/Home/Home.jsx
gonuguntla: I thought it will effect for all h1 's. I feel class is the best option

bigfishdesign13:
The "Cancel My Order" button label needs to be adjusted.  Otherwise, the updates look good.
image

PR #1388, opened 2024-06-20
Added book background color

gonuguntla:


app/javascript/components/ShowBookImage.jsx
```diff
@@ -25,7 +28,7 @@ export default function ShowBookImage(props) {
     ) {
       return (
         <Box
-          bg="var(--nypl-colors-ui-gray-x-light-cool)"
+          bg={bookBgColor}
```
bigfishdesign13, line 31: When inspecting the DOM, I see that there is a background color that is also being set on this element's parent.  You should be able to just use the styles that are applied to the parent.

In addition to that, after seeing the updated background color, we also need to update the color value for the link.  For dark mode, please set the `color` value to `dark.ui.link.primary`.

bigfishdesign13 (commented):
There are overlapping styles that can be cleaned up.  Please see my comment.

PR #1390, opened 2024-06-24
Added book titles

gonuguntla:


app/javascript/components/ShowBookImage.jsx
```diff
@@ -28,11 +29,11 @@ export default function ShowBookImage(props) {
     ) {
       return (
         <Box
-          bg={bookBgColor}
           paddingLeft="s"
           padding="s"
           width="189px"
           height="189px"
+          color={bookTitleColor}
```
bigfishdesign13, line 36: Can you incude the `color` attribute with the same set of styles that set the background color for the `<a>` element?

app/javascript/components/TeacherSetDetails.jsx
```diff
@@ -644,17 +644,19 @@ export default function TeacherSetDetails(props) {
...
-            <Link 
-              id={"ts-books-" + i}
-              onClick={(e) => {
-                windowScrollTop(e, book);
-              }}>
-              {book.title}
-            </Link>
-          </li>
-        ))}
+        { books.map((book, i) => (
```
bigfishdesign13, line 647: If there are zero book titles for a teacher set, can you hide the `Book titles` row?  For example, when the teacher set is a game.

PR #1447, opened 2024-10-10
Supporting multiple bibs in MLN bib api's (Post & delete)

gonuguntla:


gonuguntla:
Need's to update swagger document in platform docs and local env.

nonword (approved):
Seems like that should do it. The app will now accept multiple bibs in the request body and respond with a response payload that includes information about the result of each bib update. The http status will be 500 if any of the writes encounter a database, elasticsearch, or other runtime error, which should cause the mln-bib-poster to retry. If no 5xx errors occur with any bib update, the http status will be 200.

PR #1464, opened 2024-11-21
Elasticsearch update 6.8 to 7.10

gonuguntla:


docker-compose.yml
```diff
@@ -1,5 +1,7 @@
 services:
   webapp:
+    stdin_open: true
```
gonuguntla, line 3: This is I updated here for local debugging

croyfish (approved):
These changes look good to me, but we need to wait until the QA ElasticSearch instance has been upgraded to merge because tiger-team-feature deploys automatically now.

PR #1549, opened 2025-10-30
[SWIS-149] Migrate to Syndetics for cover images

russellcullen:
This moves us to using Syndetics for book covers, removing the now deprecated ContentCafe links. 

Ticket: [SWIS-149](https://newyorkpubliclibrary.atlassian.net/browse/SWIS-149)

[SWIS-149]: https://newyorkpubliclibrary.atlassian.net/browse/SWIS-149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/book.rb
+++ b/app/models/book.rb
@@ -89,7 +89,7 @@ class Book < ActiveRecord::Base
   def cover_uri
     return nil if isbn.nil?
 
-    "https://secure.syndetics.com/index.aspx?isbn=#{isbn}/MC.gif&client=nyplvega&type=hw7"
+    "https://secure.syndetics.com/index.aspx?isbn=#{isbn}/LC.gif&client=nyplvega&type=hw7"
   end
 
   # def update_from_catalog_item(item)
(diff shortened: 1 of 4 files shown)


app/models/book.rb
```diff
@@ -77,15 +77,20 @@ def self.ransackable_attributes(auth_object = nil)
...
 
-    deriv = 'S'
-    deriv = 'M' if size == :medium
-    deriv = 'L' if size == :large
+    deriv = 'SC'
+    deriv = 'MC' if size == :medium
+    deriv = 'LC' if size == :large
 
-    cover_uri.sub /Type=L/, "Type=#{deriv}"
+    "https://secure.syndetics.com/index.aspx?isbn=#{isbn}/#{deriv}.gif&client=nyplvega&type=hw7"
```
roseb89, line 86: Can we get the test data for the {isbn} Why the change to the derivatives? from 'S 'to 'SC', etc? Did the business requirements change around the new derivatives?

russellcullen: `isbn` is the publication ISBN number, I don't know if we have any particular ISBNs that we test with, but the few ISBNs I looked at seemed to work as intended. 

The cover provider has changed from ContentCafe to Syndetics, which uses a different schema and that's why we need to make these changes. 

Here are the new Syndetics docs referenced in the ticket -  https://www.librarything.com/syndeticsunbound/GetStarted_SyndeticsUnbound.pdf
Page 21 is where the URL is defined

roseb89: Well hopefully this ISBN works https://www.mylibrarynyc.org/book_details/2360

Yes, it does work.  https://secure.syndetics.com/index.aspx?isbn=9780312661052/SC.gif&client=nyplvega&type=hw7

Screenshot 2025-10-30 at 3 16 48 PM

Screenshot 2025-10-30 at 3 17 57 PM
Screenshot 2025-10-30 at 3 18 09 PM

russellcullen: Yay! You can also check out MLN QA for other usages as this change has been deployed there for validation.

PR #1556, opened 2025-12-15
Production deploy

russellcullen:
Two main changes in this release:
- Remove Adobe Analytics
- Add December 2025 school


data/public/Dec_2025_update_school_zcodes_in_mln.csv
```diff
@@ -0,0 +1,2 @@
+LOCATION_CODE,LOCATION_NAME,PRIMARY_ADDRESS_LINE_1,STATE_CODE,PRINCIPAL_PHONE_NUMBER,Location 1
+Q841,P.S. Q023 @ 841,48-25 37 Street,NY,(718) 610-3154,Queens 11101
```
Toxiapo, line 2: Will these line breaks cause problems?

Toxiapo (approved):
LGTM!

data/public/Dec_2025_update_school_zcodes_in_mln.csv
russellcullen: I ran the script locally and in QA without problem -- so I don't think so. It's a good question though, Im actually not quite sure if the csv "spec" cares about ending newlines, maybe it just comes down to how parsing is implemented in the ruby script 🤷

